Issue #1, opened 2026-03-20
Integration of GPD


FaroutYLq:
We want to explore how would [GPD](https://github.com/psi-oss/get-physics-done) help us especially in formulating fitting and literature review.

FaroutYLq:
## Analysis: Integrating GPD into mtf

> **TL;DR** — GPD's MCP servers are the right integration point. They turn mtf's passive text-generating agents into *active physics verifiers* that call structured tools during their reasoning loop, shifting the selection criterion from chi² toward physical correctness.

---

### What GPD Is

[Get Physics Done](https://github.com/psi-oss/get-physics-done) is a runtime-agnostic agentic physics framework by Physical Superintelligence PBC. Its most relevant feature for mtf is **six standalone MCP servers** (FastMCP over stdio, installable via `pip install get-physics-done`) that expose callable physics tools:

| Server | Key Tools |
|---|---|
| `gpd-verification` | `get_checklist(domain)`, `run_check(check_id, domain, artifact)`, `dimensional_check`, `limiting_case_check`, `symmetry_check` |
| `gpd-errors` | `check_error_classes(computation_desc)` → top-15 relevant error classes from a catalog of 104, `get_detection_strategy(error_id)` |
| `gpd-protocols` | `route_protocol(computation_type)` → matches free-text to 47+ domain protocols, `get_protocol(name)` → full step-by-step methodology with checkpoints |
| `gpd-conventions` | `subfield_defaults(domain)` → canonical convention defaults for 18 physics subfields, `convention_diff(lock_a, lock_b)` |
| `gpd-patterns` | `lookup_pattern(domain, category, keywords)`, `add_pattern(...)`, `promote_pattern(...)` — persistent cross-session error pattern library stored in `~/.gpd/` |
| `gpd-state` | Project state management (less relevant for mtf's session-scoped design) |

---

### The Core Problem GPD Solves in mtf

Three places in the current code where chi² implicitly wins over physical correctness:

1. **`agents/fitting.py:58-59`** — `FittingAgent.fit()` explicitly asks for `'chi_squared'`, `'reduced_chi_squared'` as mandatory result fields. No `'physical_validity'`, no `'limiting_cases'`.

2. **`debate.py:37-39`** — `DebateEngine.synthesize()` says "highlighting the strongest hypotheses." It never defines *strongest* — in fitting context the model infers chi².

[...]

PR #2, opened 2026-03-20
Integrate GPD MCP servers for physics-first hypothesis ranking

FaroutYLq:
## What this PR does

Integrates [Get Physics Done (GPD)](https://github.com/psi-oss/get-physics-done) MCP servers into mtf's agent pipeline so that hypothesis selection is driven by **physical correctness** rather than chi-squared alone.

**Design principle**: use GPD's existing tools rather than reimplementing physics verification. GPD already ships 104 curated error classes, step-by-step protocols for 47+ domains, convention databases for 18 subfields, and structured verification checks — all exposed as callable MCP tools. mtf consumes them at runtime through a bridge layer.

Closes / addresses #1.

---

## Architecture overview

```
Orchestrator
  │
  ├─ GPDMCPClient.start()          5 MCP server subprocesses
  ├─ seed_patterns()               bootstrap ~/.gpd/ error pattern library
  │
  ├─ Literature Phase
  │   ├─ subfield_defaults() × N domains  →  MemoryKind.CONVENTIONS
  │   └─ LiteratureAgent × N
  │       ├─ [GPD tool] check_error_classes()   flag error-prone hypotheses
  │       └─ [GPD tool] route_protocol()        identify relevant methodology
  │
  ├─ Fitting Phase
  │   └─ FittingAgent × N×M
  │       ├─ [GPD tool] route_protocol()        find canonical methodology
  │       ├─ [GPD tool] get_protocol()          retrieve steps + checkpoints
  │       └─ [GPD tool] subfield_defaults()     correct conventions
  │           result dict gains: protocol_followed, physical_parameter_ranges,
  │                              protocol_checkpoints_satisfied
  │
  ├─ Review Phase
  │   └─ ReviewerAgent × K
  │       ├─ [GPD tool] check_error_classes()   surface known errors
  │       ├─ [GPD tool] get_checklist()         domain checklist
  │       ├─ [GPD tool] run_check(5.1/5.2/5.3/5.18)  structured PASS/FAIL
  │       ├─ [GPD tool] lookup_pattern()        cross-session recall
  │       └─ [GPD tool] add_pattern()           persist new errors
  │           output: SUPPORTED/PLAUSIBLE/SPECULATIVE/REJECTED + check IDs
  │
  └─ DebateEngine.synthesize()
      physics-first ranking: checks > parsimony > first-principles > chi²
      context includes: CONVENTIONS + PHYSICS_VERDICT entries
```

---

### New files
- **`mtf/tools/gpd_mcp.py`** — `GPDMCPClient` bridge. Runs GPD MCP servers as subprocesses in a dedicated background thread with its own event loop. `make_tool()` wraps MCP tools as `sdk.Tool` objects with proper `inspect.Signature` for JSON schema generation. `async_call()` offloads blocking waits to thread pool. Per-session `threading.Lock` for concurrent safety. Deferred thread creation (side-effect-free `__init__`).
- **`tests/test_gpd_integration.py`** — 14 test cases covering GPDMCPClient lifecycle, MemoryKind additions, agent backward compatibility, debate engine phase-conditional prompts, config defaults, and multi-domain config. All tests work without `get-physics-done` installed.

### Modified files
- **`mtf/config.py`** — Added `enable_gpd_mcp: bool`, `physics_domains: list[str]`, `gpd_servers: list[str]`
- **`mtf/memory.py`** — Added `CONVENTIONS` and `PHYSICS_VERDICT` to `MemoryKind`
- **`mtf/orchestrator.py`** — Creates/starts/closes `GPDMCPClient`; seeds patterns at startup; passes `gpd=` to all phases
- **`mtf/debate.py`** — Physics-first ranking criterion for fitting/review phases; injects CONVENTIONS and PHYSICS_VERDICT into synthesis context
- **`mtf/phases/literature_phase.py`** — Locks conventions per domain before fan-out; passes 2 GPD tools to LiteratureAgents
- **`mtf/phases/fitting_phase.py`** — Passes 3 GPD tools to all FittingAgents
- **`mtf/phases/review_phase.py`** — Passes 8 GPD tools to all ReviewerAgents
- **`mtf/agents/literature.py`** — Accepts `gpd_tools`; enriched prompt classifies hypotheses by physical basis
- **`mtf/agents/fitting.py`** — Accepts `gpd_tools`; enriched prompt follows GPD protocols and reports physical parameter ranges
- **`mtf/agents/reviewer.py`** — Accepts `gpd_tools`; enriched prompt runs structured verification checks and produces verdicts with check IDs
- **`mtf/cli.py`** — Added `--physics-domains`, `--no-gpd`, `--gpd-servers` flags
- **`pyproject.toml`** — Added `[gpd]` optional extra (`get-physics-done>=1.1.0`, `mcp>=1.26.0`)
- **`README.md`** — Mermaid diagram updated with GPD servers; agents table updated; new "GPD Physics Verification" section; installation/CLI/SharedMemory sections updated
- **`CLAUDE.md`** — GPD section rewritten with design-principle table; tool flow per agent; multi-domain docs; guidance to check GPD before reimplementing

---

## Why GPD servers instead of reimplementing

| Capability | GPD provides | What mtf would have to build without GPD |
|---|---|---|
| Physics error detection | 104 curated error classes with detection strategies | Encode error knowledge in system prompts per domain |
| Computation protocols | 47+ domain protocols with checkpoints | Write step-by-step instructions per domain in prompts |
| Convention locking | 18 standard fields across 14 subdomains | Track conventions manually in SharedMemory |
| Verification checks | Automated check IDs 5.1–5.19 with domain-specific issue detection | Build a custom verification framework |
| Cross-session memory | `~/.gpd/` pattern library with confidence promotion | Build a persistent store (database, filesystem) |

---

## Multi-domain support

`physics_domains` is a `list[str]` (not a single string), so cross-domain phenomena get conventions and checklists from all relevant domains:

```bash
mtf --physics-domains condensed_matter qft "topological insulator edge states"
mtf --physics-domains gr nuclear amo "neutron star cooling anomaly"
```

Convention locking iterates over all domains, storing one `MemoryKind.CONVENTIONS` entry per domain. ReviewerAgent's system prompt instructs `get_checklist` to be called once per domain and merge the results.

---

## Backward compatibility

- All agents: `gpd_tools=None` (default) preserves original behavior
- All phases: `gpd=None` (default) preserves original behavior
- `enable_gpd_mcp=True` but GPD not installed → graceful degradation, warning logged, pipeline runs normally
- No existing test behavior changed

---

## Test plan
- [ ] `pytest tests/test_gpd_integration.py -v` passes
- [ ] `pytest tests/test_memory.py tests/test_debate.py tests/test_phases.py -v` still passes
- [ ] `mtf --help` shows `--physics-domains`, `--no-gpd`, `--gpd-servers`
- [ ] `mtf --no-gpd "test"` runs without GPD
- [ ] `mtf --physics-domains qft gr "test"` locks conventions for both domains

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,24 @@ Each phase: fan-out agents with `asyncio.gather()` → collect reports → `Deba
 ### Concurrency
 Fitting agents are rate-limited by `asyncio.Semaphore(config.fitting_semaphore_limit)` (default 6) to avoid overwhelming the API when `fitting_scope="per_hypothesis"` spawns `N_hypotheses × M` concurrent agents.
 
+### GPD MCP Integration
+MTF optionally connects to [get-physics-done](https://github.com/FaroutYLq/get-physics-done) MCP servers for physics verification. Install with `pip install -e ".[dev,gpd]"`. Controlled by `config.enable_gpd_mcp` (default `True`; no-ops gracefully if the package is missing).
+
+**Servers** (defined in `mtf/tools/gpd_mcp.py`):
+- `verification` — structured physics checks (dimensional analysis 5.1, symmetry 5.2, limiting cases 5.3, fit-family mismatch 5.18)
+- `errors` — catalog of 104 known physics error classes with detection strategies
+- `protocols` — canonical computation protocols (step-by-step methodology with checkpoints)
+- `conventions` — subfield-specific sign conventions, Fourier transforms, natural units
+- `patterns` — persistent cross-session library of discovered physics error patterns
+
+**`GPDMCPClient`** runs a dedicated background event loop in a daemon thread. Each server is a subprocess communicating over stdio MCP protocol. `make_tool(server, tool_name, description)` returns an `sdk.Tool` (or `None` if unavailable), which phases filter into agent tool lists. `call(server, tool_name, **kwargs)` provides synchronous access for one-shot calls (e.g. seeding patterns at startup).
+
+**New `MemoryKind` values**:
+- `CONVENTIONS` — physics convention snapshot locked at the start of the literature phase; included in all agent prompt contexts
+- `PHYSICS_VERDICT` — structured check results from the verification server; injected into debate synthesis context
+
+**Pipeline flow**: conventions are locked once in the literature phase via `subfield_defaults`. All three agent types accept `gpd_tools: list | None` for backward compatibility. `DebateEngine.synthesize()` conditionally adds a physics-first ranking criterion to the system prompt for fitting and review phases (not literature). Conventions and physics verdicts from memory are appended to the user content block sent to the synthesis call.
+
 ## Key Invariants
 
 - All `HumanInterface` methods are async; never call `input()` directly.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,8 @@ mtf/
 ├── tools/
 │   ├── arxiv_search.py     sdk.Tool wrapping arxiv.Client
 │   ├── semantic_search.py  sdk.Tool wrapping semanticscholar API
-│   └── fitting_tools.py    run_fitting_code() — exec-based sandboxed runner
+│   ├── fitting_tools.py    run_fitting_code() — exec-based sandboxed runner
+│   └── gpd_mcp.py          GPDMCPClient — bridges GPD MCP servers to sdk.Tool
 └── toolkit/
     └── registry.py         ToolkitRegistry — user-provided data arrays + model callables
 ```
@@ -51,6 +52,9 @@ mtf/
 ```bash
 # Install (editable + dev extras)
 pip install -e ".[dev]"
+
+# Install with GPD physics verification
+pip install -e ".[dev,gpd]"
 
 # Lint
 ruff check mtf tests
@@ -93,22 +97,35 @@ Each phase: fan-out agents with `asyncio.gather()` → collect reports → `Deba
 Fitting agents are rate-limited by `asyncio.Semaphore(config.fitting_semaphore_limit)` (default 6) to avoid overwhelming the API when `fitting_scope="per_hypothesis"` spawns `N_hypotheses × M` concurrent agents.
 
 ### GPD MCP Integration
-MTF optionally connects to [get-physics-done](https://github.com/FaroutYLq/get-physics-done) MCP servers for physics verification. Install with `pip install -e ".[dev,gpd]"`. Controlled by `config.enable_gpd_mcp` (default `True`; no-ops gracefully if the package is missing).
 
-**Servers** (defined in `mtf/tools/gpd_mcp.py`):
-- `verification` — structured physics checks (dimensional analysis 5.1, symmetry 5.2, limiting cases 5.3, fit-family mismatch 5.18)
-- `errors` — catalog of 104 known physics error classes with detection strategies
-- `protocols` — canonical computation protocols (step-by-step methodology with checkpoints)
-- `conventions` — subfield-specific sign conventions, Fourier transforms, natural units
-- `patterns` — persistent cross-session library of discovered physics error patterns
+**Design principle: use existing GPD tools rather than reimplementing physics verification.** GPD ([Get Physics Done](https://github.com/psi-oss/get-physics-done)) already ships 104 curated error classes, step-by-step protocols for 47+ physics domains, convention databases for 18 subfields, and structured verification checks — all exposed as callable MCP tools. mtf consumes these tools at runtime through a bridge layer rather than duplicating this physics knowledge in its own prompts or code.
 
-**`GPDMCPClient`** runs a dedicated background event loop in a daemon thread. Each server is a subprocess communicating over stdio MCP protocol. `make_tool(server, tool_name, description)` returns an `sdk.Tool` (or `None` if unavailable), which phases filter into agent tool lists. `call(server, tool_name, **kwargs)` provides synchronous access for one-shot calls (e.g. seeding patterns at startup).
+Install with `pip install -e ".[dev,gpd]"`. Controlled by `config.enable_gpd_mcp` (default `True`; no-ops gracefully if the package is missing).
 
-**New `MemoryKind` values**:
-- `CONVENTIONS` — physics convention snapshot locked at the start of the literature phase; included in all agent prompt contexts
+**Servers** (managed by `GPDMCPClient` in `mtf/tools/gpd_mcp.py`):
+
+| Server | GPD tools used by mtf | Why mtf uses GPD instead of reimplementing |
+|---|---|---|
+| `verification` | `get_checklist`, `run_check`, `dimensional_check`, `limiting_case_check` | GPD maintains domain-specific checklists (check IDs 5.1–5.19) with automated issue detection. Reimplementing would require encoding physics knowledge per domain. |
+| `errors` | `check_error_classes`, `get_detection_strategy` | GPD's catalog of 104 error classes (sign errors, missing factors of 2π, gauge artifacts, etc.) with detection strategies is a curated knowledge base that improves over time. |
+| `protocols` | `route_protocol`, `get_protocol` | GPD provides canonical step-by-step methodology with checkpoints for each physics domain. FittingAgent follows these instead of inventing ad-hoc procedures. |
+| `conventions` | `subfield_defaults`, `convention_check` | GPD tracks 18 standard convention fields (metric signature, Fourier convention, natural units, gauge choice, etc.) across 14 subdomains. Prevents silent convention mismatches between agents. |
+| `patterns` | `lookup_pattern`, `add_pattern`, `seed_patterns` | GPD's `~/.gpd/` pattern store is the only persistent cross-session memory in the mtf pipeline. Errors found in one run surface in future runs on the same domain. |
+
+**`GPDMCPClient`** runs a dedicated background event loop in a daemon thread. Each server is a subprocess communicating over stdio MCP protocol. `make_tool(server, tool_name, description, params=)` returns an `sdk.Tool` (or `None` if unavailable), which phases filter into agent tool lists. `call(server, tool_name, **kwargs)` provides synchronous access for phase-level one-shot calls (e.g. convention locking, seeding patterns). `async_call()` offloads the blocking wait to a thread pool for use inside `asyncio.gather()` fan-outs.
+
+**How GPD tools flow to each agent**:
+- `LiteratureAgent` receives: `check_error_classes` (flag error-prone hypotheses), `route_protocol` (identify correct methodology)
+- `FittingAgent` receives: `route_protocol` + `get_protocol` (follow canonical procedure), `subfield_defaults` (correct conventions)
+- `ReviewerAgent` receives: all 8 tools above — `get_checklist`, `run_check` (5.1/5.2/5.3/5.18), `dimensional_check`, `limiting_case_check`, `check_error_classes`, `get_detection_strategy`, `lookup_pattern`, `add_pattern`
+
+**`MemoryKind` values for GPD data**:
+- `CONVENTIONS` — physics convention snapshot locked per domain at the start of the literature phase; included in all agent prompt contexts
 - `PHYSICS_VERDICT` — structured check results from the verification server; injected into debate synthesis context
 
-**Pipeline flow**: conventions are locked once in the literature phase via `subfield_defaults`. All three agent types accept `gpd_tools: list | None` for backward compatibility. `DebateEngine.synthesize()` conditionally adds a physics-first ranking criterion to the system prompt for fitting and review phases (not literature). Conventions and physics verdicts from memory are appended to the user content block sent to the synthesis call.
+**Pipeline flow**: conventions are locked once in the literature phase via `subfield_defaults` (called once per domain in `config.physics_domains`). All three agent types accept `gpd_tools: list | None` for backward compatibility. `DebateEngine.synthesize()` conditionally adds a physics-first ranking criterion to the system prompt for fitting and review phases (not literature). Conventions and physics verdicts from memory are appended to the user content block sent to the synthesis call.
+
+**When adding new physics capabilities to mtf**: check whether GPD already provides it as an MCP tool before implementing from scratch. The `gpd-mcp-skills` server (`list_skills`, `route_skill`) can discover available GPD capabilities programmatically.
 
 ## Key Invariants
 
--- a/README.md
+++ b/README.md
@@ -10,30 +10,44 @@ A multi-agent AI system for experimental physicists. Describe an unexplained phe
 flowchart TD
     Input(["📋 User Input\nphenomenon description + images + toolkit data"])
 
+    subgraph GPD ["🔧 GPD MCP SERVERS (optional)"]
+        direction TB
+        GV["verification\nchecks 5.1–5.19"]
+        GE["errors\n104 error classes"]
+        GP["protocols\n47+ domain protocols"]
+        GC["conventions\n18 subfields"]
+        GPat["patterns\ncross-session memory"]
+    end
+
     subgraph LIT ["① LITERATURE PHASE"]
         direction TB
-        L["L1 · L2 · L3\nN parallel agents\narxiv + Semantic Scholar"]
+        LC["Lock conventions\nvia GPD subfield_defaults"]
+        L["L1 · L2 · L3\nN parallel agents\narxiv + Semantic Scholar\n+ GPD: check_error_classes, route_protocol"]
         LD["🔀 Debate\nsynthesis call"]
         LU{"User approval"}
-        L --> LD --> LU
+        LC --> L --> LD --> LU
         LU -->|"reject: add feedback"| L
     end
 
     subgraph FIT ["② FITTING PHASE  —  per approved hypothesis"]
         direction TB
         FT["toolkit check\n(request missing data from user)"]
-        F["F1 · F2 · F3\nM parallel agents\nlmfit + numpy/scipy"]
-        FD["🔀 Debate\nsynthesis call"]
+        F["F1 · F2 · F3\nM parallel agents\nlmfit + numpy/scipy\n+ GPD: route_protocol, get_protocol, subfield_defaults"]
+        FD["🔀 Debate\nphysics-first ranking"]
         FU{"User approval"}
         FT --> F --> FD --> FU
     end
 
     subgraph REV ["③ REVIEW PHASE"]
         direction TB
-        R["R1 · R2 · R3\nK parallel agents\ntheory validity + next experiments"]
-        RD["🔀 Debate\nfinal synthesis"]
+        R["R1 · R2 · R3\nK parallel agents\n+ GPD: get_checklist, run_check, check_error_classes,\nlookup_pattern, add_pattern"]
+        RD["🔀 Debate\nphysics-first ranking"]
         R --> RD
     end
+
+    GPD -.->|"tools"| L
+    GPD -.->|"tools"| F
+    GPD -.->|"tools"| R
 
     Report(["📄 Final Report"])
 
@@ -60,9 +74,9 @@ Images are digested first (phase ⓪): `ImageDigestAgent` uses Claude's vision A
 | Agent | Phase | API | Tools | Memory written | Description |
 |-------|-------|-----|-------|----------------|-------------|
 | `ImageDigestAgent` | ⓪ Pre-processing | `messages.create()` (multimodal) | — | `IMAGE_DATA` | Encodes each user image as base64 and calls the vision API to extract plot type, axis labels/units/scale, numerical data series, quantitative features, and annotations. Runs in parallel, one instance per image. |
-| `LiteratureAgent` | ① Literature | `sdk.query()` | arxiv search, Semantic Scholar | `LITERATURE` | Searches arxiv and Semantic Scholar for papers relevant to the phenomenon. Returns a structured report: phenomenon summary, cited papers, ranked hypotheses, and key equations. Spawns N parallel instances (default 3). |
-| `FittingAgent` | ② Fitting | `sdk.query()` | `run_fitting_code()` (exec sandbox) | `FIT_RESULT` | Given an approved hypothesis and toolkit data, writes and executes `lmfit`/`numpy`/`scipy` fitting code. Reports χ², reduced χ², best-fit parameters with uncertainties, and hypothesis support. Rate-limited by `asyncio.Semaphore`. |
-| `ReviewerAgent` | ③ Review | `sdk.query()` | — | `REVIEW` | Evaluates hypothesis validity against data and literature, identifies weaknesses and confounds, and proposes specific follow-up experiments to distinguish competing explanations. Spawns K parallel instances (default 3). |
+| `LiteratureAgent` | ① Literature | `sdk.query()` | arxiv search, Semantic Scholar, GPD: `check_error_classes`, `route_protocol` | `LITERATURE` | Searches arxiv and Semantic Scholar. Calls GPD to identify error-prone hypotheses and the relevant computation protocol. Classifies each hypothesis by physical basis (first-principles / semi-empirical / purely empirical). Spawns N parallel instances (default 3). |
+| `FittingAgent` | ② Fitting | `sdk.query()` | GPD: `route_protocol`, `get_protocol`, `subfield_defaults` | `FIT_RESULT` | Given an approved hypothesis and toolkit data, retrieves the canonical domain protocol from GPD and writes fitting code that follows its checkpoints. Reports χ², parameters, and protocol compliance. Rate-limited by `asyncio.Semaphore`. |
+| `ReviewerAgent` | ③ Review | `sdk.query()` | GPD: `get_checklist`, `run_check`, `dimensional_check`, `limiting_case_check`, `check_error_classes`, `get_detection_strategy`, `lookup_pattern`, `add_pattern` | `REVIEW` | Runs GPD's structured verification checks (5.1 dimensional, 5.2 symmetry, 5.3 limiting cases, 5.18 fit-family) against each fit result. Produces SUPPORTED/PLAUSIBLE/SPECULATIVE/REJECTED verdicts citing check IDs. Records new error patterns for future sessions. Spawns K parallel instances (default 3). |
 | `ToolBuilderAgent` | ② Fitting (on demand) | `sdk.query()` | — | `TOOLKIT_DIGEST` | Invoked when a fitting agent requests toolkit data that the user has not pre-registered. Parses raw user input (functions, CSVs, code snippets) into `data_items` and `model_items` dicts via LLM-generated `exec()` code, then registers them in `ToolkitRegistry`. |
 
 All agents except `ImageDigestAgent` extend `BaseAgent`, which prepends a formatted `SharedMemory` context block before every `sdk.query()` call so every agent sees prior debate summaries, image digests, and user feedback.
@@ -78,13 +92,16 @@ All agents except `ImageDigestAgent` extend `BaseAgent`, which prepends a format
 git clone https://github.com/your-org/mtf.git
 cd mtf
 
-# Option A — pip
+# Option A — pip (without GPD physics verification)
 pip install -e ".[dev]"
+
+# Option A — pip (with GPD physics verification — recommended)
+pip install -e ".[dev,gpd]"
 
 # Option B — conda
 conda env create -f environment.yml
 conda activate mtf
-pip install -e .
+pip install -e ".[gpd]"
 
 # Set your API key
 export ANTHROPIC_API_KEY="sk-ant-..."
@@ -126,6 +143,9 @@ mtf
 | `--debate-model` | `claude-opus-4-6` | Model for debate synthesis |
 | `--images` | _(none)_ | Image files to digest (PNG, JPG, GIF, WebP; space-separated) |
 | `--image-digest-model` | `claude-opus-4-6` | Model for image digestion |
+| `--physics-domains` | `condensed_matter` | Physics domains for GPD (space-separated, e.g. `condensed_matter qft`) |
+| `--no-gpd` | _(off)_ | Disable GPD MCP physics verification servers |
+| `--gpd-servers` | _(all)_ | Which GPD servers to start (verification, errors, protocols, conventions, patterns) |
 
 ### 2. Python API — with your own data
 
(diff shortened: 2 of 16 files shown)


FaroutYLq:
## Architecture Review

### 1. `**kwargs` tool functions lack typed signatures -- sdk.Tool.from_function will generate empty input schemas

**Severity:** Critical

**File:** `mtf/tools/gpd_mcp.py`, lines 145-150

The `make_tool()` closure creates `def _tool_fn(**kwargs: Any) -> str:` and passes it to `sdk.Tool.from_function()`. Compare with the existing pattern in `mtf/tools/arxiv_search.py` (line 12):

```python
def arxiv_search(query: str, max_results: int = 5) -> list[dict[str, str]]:
```

`sdk.Tool.from_function()` introspects the function signature to build the JSON schema that tells the model what parameters the tool accepts. A bare `**kwargs` produces an empty `parameters` schema, so the model has no idea what arguments to pass -- it will either pass nothing or hallucinate parameter names. This affects every single GPD tool (all 8+ tools created via `make_tool`).

**Suggested fix:** Either (a) query the MCP server's `tools/list` endpoint at startup to get the real tool schemas and dynamically build typed wrappers, or (b) accept a `parameters: dict` argument in `make_tool()` describing the expected params and construct the `sdk.Tool` manually with an explicit `input_schema` rather than relying on `from_function()` introspection.

---

### 2. Deadlock risk: sync `_run()` called from within an async context via sdk.query tool callbacks

**Severity:** Critical

**File:** `mtf/tools/gpd_mcp.py`, lines 161-164

`_tool_fn` (line 145) is a synchronous function that calls `self.call()` which calls `self._run()` which calls `future.result(timeout=60)` -- this **blocks the calling thread**. When `sdk.query()` invokes a tool function, it does so from within the main asyncio event loop's thread. Blocking the event loop thread for up to 60 seconds will freeze all other concurrent agents launched via `asyncio.gather()`.

During the literature phase, for example, `asyncio.gather()` fans out 3 `LiteratureAgent` instances (line 68 of `literature_phase.py`). If agent 1's tool call blocks the event loop for 60s waiting on the MCP subprocess, agents 2 and 3 are completely stalled.

**Suggested fix:** The tool function should be async (if `sdk.Tool` supports async callables), or the blocking `_run()` call should be wrapped in `asyncio.to_thread()` so it doesn't block the event loop. Alternatively, if `sdk.query()` already runs tool functions in a thread executor, this is fine -- but that behavior needs to be verified against the SDK documentation.

---

### 3. `GPDMCPClient.__init__` unconditionally starts a background thread and event loop

**Severity:** Warning

**File:** `mtf/tools/gpd_mcp.py`, lines 49-54; `mtf/orchestrator.py`, line 47

The constructor immediately creates a new event loop and starts a daemon thread. In `orchestrator.py` line 47:
```python
gpd = GPDMCPClient() if self._config.enable_gpd_mcp else None
```

If `start()` then fails (e.g., `mcp` not installed), the thread and loop are left running. The `close()` method stops the loop and joins the thread, but `close()` is only called in the `finally` block at line 120 -- which is fine for the happy path. However, if the user creates a `GPDMCPClient()` but never calls `start()` (or construction happens in test code), the dangling thread/loop is never cleaned up.

**Suggested fix:** Defer thread/loop creation to `start()`, or make `GPDMCPClient` a context manager so `__exit__` guarantees cleanup.

---

### 4. Partial server startup leaks subprocesses of successfully-started servers

**Severity:** Warning

**File:** `mtf/tools/gpd_mcp.py`, lines 75-100

In `_start_async`, if server 3 out of 5 fails, servers 1 and 2 have live subprocess sessions registered in `self._sessions` and managed by `self._exit_stack`. The exception is caught per-server (line 99-100), so this specific scenario is actually handled correctly for individual server failures.

However, if `_start_async` itself raises an unexpected exception *before* entering the per-server loop (e.g., the `from mcp import ClientSession` import fails), `self._exit_stack` has already been entered via `__aenter__()` (line 80) but will never be exited -- the `start()` method's except block (line 71) sets `self._available = False` but does not close the exit stack. The `close()` method will later try `self._exit_stack.__aexit__()`, which may or may not properly clean up a partially-initialized stack.

**Suggested fix:** Wrap the `_start_async` body in a try/except that calls `self._exit_stack.__aexit__()` on failure, or restructure to use `async with AsyncExitStack() as stack:` so cleanup is automatic.

---

### 5. Duplicate tool names across agents: `check_error_classes` registered in both literature and review agents

**Severity:** Warning

**File:** `mtf/phases/literature_phase.py`, line 31; `mtf/phases/review_phase.py`, line 45

Both phases create a tool named `check_error_classes` (from the "errors" server). Since each agent gets its own tool list, this is not a conflict *between agents*. However, within `review_phase.py` itself, the reviewer agent receives 8 tools from various servers. If two GPD servers happened to expose a tool with the same name, `make_tool` would create two `sdk.Tool` objects with the same `__name__`. Whether `sdk.query()` handles duplicate tool names correctly depends on the SDK internals -- it may silently shadow one.

This is currently safe because each tool name maps to a unique MCP server tool, but the architecture has no guard against it. A simple `seen_names` set check in the tool list construction would prevent subtle bugs if servers evolve.

---

### 6. Unbounded memory growth: CONVENTIONS and PHYSICS_VERDICT entries accumulate across debate rounds

**Severity:** Warning

**File:** `mtf/memory.py` (entire file); `mtf/phases/literature_phase.py`, lines 43-45; `mtf/debate.py`, lines 69-74

`SharedMemory` is append-only (line 41). In `literature_phase.py` line 44, `MemoryKind.CONVENTIONS` entries are added. If the literature phase runs multiple debate rounds (up to `max_debate_rounds`), the conventions are only added once (before the loop, line 42-48), which is fine.

However, `PHYSICS_VERDICT` entries are never explicitly added in the PR's code -- they appear in `debate.py` line 70 where they're *read* but the PR doesn't show where they're *written*. If the GPD verification tools produce verdicts that agents store via `memory.add(MemoryKind.PHYSICS_VERDICT, ...)` in future iterations, these will accumulate without bound.

More concretely, `debate.py` line 34 calls `self._memory.format_context()` with no filter (returns ALL entries). Every synthesis call dumps the entire memory into the prompt. As debate rounds accumulate LITERATURE, DEBATE, USER_FEEDBACK, CONVENTIONS entries, the prompt grows monotonically. With 3 literature agents x 3 rounds + 3 fitting agents x N hypotheses + 3 reviewers, the context can exceed model limits.

**Suggested fix:** Either (a) add a `SharedMemory.clear(kind)` method and prune stale entries between phases, or (b) filter `format_context()` to only include the most recent N entries per kind, or (c) pass explicit `kinds` to `format_context()` in `debate.py` line 34 instead of dumping everything.

---

### 7. `enable_gpd_mcp` defaults to `True` but `mcp` is an optional dependency

**Severity:** Warning

**File:** `mtf/config.py`, line 33; `pyproject.toml`, line 27

`enable_gpd_mcp: bool = True` means GPD is opt-out. But `mcp` and `get-physics-done` are in `[project.optional-dependencies] gpd`, not in the core `dependencies` list. A fresh `pip install mtf` (without `pip install mtf[gpd]`) will result in:

1. `GPDMCPClient()` is created (orchestrator.py line 47)
2. `start()` is called, which calls `_start_async()`
3. `from mcp import ClientSession` raises `ModuleNotFoundError`
4. This is caught and logged as a warning, GPD is silently disabled

While this technically works (graceful degradation), defaulting an optional feature to `True` is surprising. Users will see a warning on every run unless they install the optional dep or explicitly set `enable_gpd_mcp=False`.

**Suggested fix:** Default `enable_gpd_mcp` to `False`, or auto-detect: `enable_gpd_mcp: bool = field(default_factory=lambda: importlib.util.find_spec("mcp") is not None)`.

---

### 8. `GPDMCPClient` is not thread-safe for concurrent `call()` invocations

**Severity:** Note

**File:** `mtf/tools/gpd_mcp.py`, lines 117-131

Multiple agents fan out via `asyncio.gather()`. If the tool functions block (issue #2 above) they'd serialize. But if they're made async or run in threads (as suggested in #2's fix), multiple `call()` invocations could submit coroutines to the background loop concurrently. The MCP `ClientSession` is unlikely to be safe for concurrent `call_tool()` calls on the same session -- MCP stdio is a single-stream protocol, and interleaving requests/responses would corrupt the message framing.

**Suggested fix:** Add a per-session `asyncio.Lock` (on the background loop) to serialize MCP calls to each server. E.g.:
```python
self._locks: dict[str, asyncio.Lock] = {}
# In _start_async, after session creation:
self._locks[name] = asyncio.Lock()
# In call():
async with self._locks[server]:
    result = await session.call_tool(tool_name, kwargs)
```

---

### 9. `close()` can hang if `_run()` timeout fires during `__aexit__`

**Severity:** Note

**File:** `mtf/tools/gpd_mcp.py`, lines 102-111

`close()` calls `self._run(self._exit_stack.__aexit__(None, None, None))` which has a 60-second timeout (from `_run`). If an MCP subprocess is hung, this blocks the main thread for 60 seconds before the `TimeoutError` is caught. Then `self._loop.call_soon_threadsafe(self._loop.stop)` stops the loop, and `self._thread.join(timeout=5)` waits another 5 seconds. Worst case: 65 seconds blocking on cleanup.

This is acceptable for a CLI tool but worth documenting. If used in a server context, consider a shorter cleanup timeout.

---

### 10. `gpd` parameter typed as `Any` throughout the phase functions

**Severity:** Note

**File:** `mtf/phases/literature_phase.py`, line 21; `mtf/phases/fitting_phase.py`, line 30; `mtf/phases/review_phase.py`, line 21

All three phase functions accept `gpd: Any | None = None`. Since `GPDMCPClient` is a concrete class in this codebase, this should be typed as `GPDMCPClient | None`. Using `Any` defeats mypy's ability to catch typos in method calls (e.g., `gpd.cal(...)` instead of `gpd.call(...)` would pass type checking).

**Suggested fix:** `from mtf.tools.gpd_mcp import GPDMCPClient` and type as `gpd: GPDMCPClient | None = None`. If avoiding the import for circular-dependency reasons, use `from __future__ import annotations` with `TYPE_CHECKING`.

FaroutYLq:
## Physics Deduction Review

Reviewing the GPD MCP integration (PR #2) from the perspective of physical reasoning correctness. The integration adds structured physics verification checks, convention locking, and a rigid ranking hierarchy to the multi-agent system. While the direction is sound, several aspects of the physics deduction logic could lead to incorrect conclusions in realistic experimental scenarios.

---

### 1. Ranking hierarchy is not universally valid for experimental physics

**Severity: Critical**

The debate engine (`debate.py`, lines 43-57) enforces a rigid ranking:

> ```
> 1. Physics checks: hypotheses that pass dimensional analysis (5.1),
>    limiting cases (5.3), symmetry (5.2), and fit-family validity (5.18)
>    rank above those that fail, regardless of chi-squared.
> 2. Parsimony: prefer fewer free parameters at comparable fit quality.
> 3. First-principles basis: a derivation from known theory outranks a
>    phenomenological ansatz at equal fit quality.
> 4. Fit quality (chi-squared, reduced chi-squared): used as tiebreaker only.
> ```

This ordering is problematic in several well-known experimental scenarios:

**(a) chi-squared as primary discriminator.** When all competing models satisfy dimensional analysis, symmetry, and limiting cases equally (which is common when candidates are drawn from the same theoretical framework), chi-squared and information criteria (AIC, BIC) become the *primary* discriminators, not tiebreakers. Example: fitting neutron scattering data to distinguish between different magnetic ordering models in a frustrated magnet --- all candidates are spin Hamiltonians with correct dimensions, symmetry, and limits, but only goodness-of-fit distinguishes them.

**(b) Parsimony vs. physical meaning.** Ranking parsimony above first-principles basis (items 2 vs. 3) means a 2-parameter empirical power law would be preferred over a 5-parameter Ginzburg-Landau free energy expansion near a phase transition, even though each GL parameter has clear physical meaning (order parameter coupling, gradient energy, external field coupling, etc.) and the power law has none. In condensed matter, the "more parameters" model is frequently the correct physical description.

**(c) "Regardless of chi-squared" is too absolute.** The text says models passing physics checks rank above those that fail "regardless of chi-squared." But a model failing a limiting-case check by 2% in a crossover regime while fitting data with chi-squared/dof = 1.01 is almost certainly better than a model passing all checks with chi-squared/dof = 15. The system should allow quantitative severity weighting of check failures rather than treating them as binary pass/fail gates.

**Suggestion:** Replace the rigid hierarchy with a weighted scoring scheme, or at minimum add language like "unless all competing models pass the same checks, in which case fit quality and information criteria become the primary ranking criterion." Consider adding AIC/BIC to the required fit outputs alongside chi-squared.

---

### 2. First-principles bias systematically disadvantages large areas of physics

**Severity: Critical**

The LiteratureAgent prompt (`literature.py`, line 33) states:

> `Prefer first-principles hypotheses over phenomenological curve fits.`

And the debate engine ranks "first-principles basis" above fit quality. This creates a systematic bias against correct answers in many subfields:

- **Condensed matter:** Ginzburg-Landau theory, Fermi liquid theory, BCS theory (originally phenomenological), Landau damping --- these are effective theories that are *the* correct framework for their respective phenomena. A user investigating a superconducting transition would have GL theory deprioritized in favor of a microscopic Hamiltonian that may be intractable or inapplicable near Tc.

- **Nuclear physics:** The liquid drop model and nuclear shell model are phenomenological yet remain the standard predictive frameworks. Ab initio nuclear structure calculations from QCD are not yet competitive for most nuclei.

- **Astrophysics:** The Tully-Fisher relation, the fundamental plane, mass-metallicity relation --- these empirical scaling laws are often more reliable predictors than first-principles models that require uncertain input physics (star formation prescriptions, feedback models).

- **Plasma physics:** Transport coefficients are routinely described by Braginskii's semi-empirical expressions rather than first-principles kinetic theory, which is often intractable.

The instruction should be reframed: "Classify each hypothesis as first-principles, semi-empirical, or purely empirical" (which the PR already does in item 4c) but *do not* use this classification as a ranking criterion. The classification is informative metadata, not a quality metric.

---

### 3. Mandatory check set is simultaneously incomplete and potentially misleading

**Severity: Warning**

The reviewer prompt (`reviewer.py`, lines 19-24) mandates checks 5.1 (dimensional), 5.2 (symmetry), 5.3 (limiting cases), and 5.18 (fit family mismatch) for *every* hypothesis. Issues:

**(a) Missing mandatory checks.** Conservation law verification (5.4) and numerical convergence (5.5) are available in `run_check` (per the review_phase.py tool description, line 35) but are not mandatory. Conservation laws are arguably more fundamental than symmetry checks --- a model that violates energy conservation is wrong regardless of symmetry properties. Numerical convergence is critical for any iterative fitting: a fit that hasn't converged has meaningless parameter values.

**(b) Symmetry check on broken-symmetry systems.** Mandating symmetry check 5.2 for every hypothesis is inappropriate when the phenomenon *is* symmetry breaking. Example: a user investigating a ferromagnetic phase transition would have hypotheses that correctly predict spontaneous magnetization (breaking rotational symmetry) flagged by a symmetry check. The check needs context about whether the system is above or below a transition.

**(c) Limiting-case check on interpolating models.** Check 5.3 asks "does the model recover known limits?" But many physical models are specifically designed for intermediate regimes where neither limit applies. Example: the BCS-BEC crossover in ultracold Fermi gases --- the correct description at unitarity does not reduce cleanly to either the BCS or BEC limit. Flagging such a model for "failing" limiting cases would be incorrect.

**(d) Dimensional analysis is often trivially satisfied.** When all competing models use the same units and are expressed in standard form, check 5.1 passes for everything and provides zero discriminating power. The check is important as a sanity filter but should not be weighted equally with more informative checks.

**Suggestion:** Make 5.4 (conservation) and 5.5 (convergence) mandatory. Add context-awareness to 5.2: the system prompt should instruct the reviewer to distinguish between "model violates a symmetry that should hold" vs. "model correctly describes symmetry breaking." Consider a "relevance" pre-filter where the reviewer decides which checks are informative for the specific phenomenon before running them.

---

### 4. Convention locking before phenomenon understanding creates domain mismatch risk

**Severity: Warning**

In `literature_phase.py` (lines 42-49), conventions are locked at the very start of the literature phase based on `config.physics_domain`:

```python
# Lock physics conventions before the first fan-out
if gpd is not None:
    conventions = gpd.call("conventions", "subfield_defaults", domain=config.physics_domain)
```

Problems:

**(a) Premature commitment.** The conventions are locked *before* any literature search has been performed. At this point, the system does not yet know what the phenomenon actually is --- only the user's initial description is available. If the user says "I see an anomalous signal in my detector" and sets `physics_domain="condensed_matter"`, but the correct explanation turns out to involve particle physics (dark matter scattering in a crystal detector, as in direct detection experiments), the locked condensed matter conventions (e.g., Fourier transform sign conventions, natural units) will be wrong for the particle physics hypothesis.

**(b) Cross-domain phenomena.** Many real experimental phenomena sit at domain boundaries:
  - Astroparticle physics (particle + astrophysics conventions conflict: natural units vs. CGS/Gaussian, $c=1$ vs. explicit $c$)
  - Quantum materials + AMO (condensed matter band structure conventions vs. atomic physics angular momentum coupling conventions)
  - Biophysics (statistical mechanics conventions vs. biochemistry conventions for free energies: sign of $\Delta G$, reference states)

A single `physics_domain: str` cannot represent these situations. There is currently no mechanism to revise conventions if literature search reveals the phenomenon spans multiple domains.

**(c) No convention conflict detection.** If a fitting agent calls `subfield_defaults` with a different subfield than what was locked, there is no mechanism to detect or resolve the conflict. The locked conventions in memory and the freshly retrieved conventions could contradict each other silently.

**Suggestion:** Defer convention locking to after the first literature debate round, when the system has identified candidate hypotheses and their domains. Support a list of domains (`physics_domains: list[str]`) and produce a merged/annotated convention set. Add a convention conflict check when multiple domains are active.

---

### 5. Verdict categories lack "INCONCLUSIVE" and "CONDITIONALLY SUPPORTED"

**Severity: Warning**

The reviewer prompt (`reviewer.py`, lines 35-36) restricts verdicts to:

> `SUPPORTED / PLAUSIBLE / SPECULATIVE / REJECTED`

Missing categories:

- **INCONCLUSIVE:** When the available data genuinely cannot distinguish between hypotheses. This is extremely common in experimental physics --- you often need a specific follow-up measurement to break a degeneracy. Forcing the system to pick among the four existing categories when the honest answer is "we cannot tell" leads to overconfident conclusions. Example: two models with chi-squared/dof of 1.02 and 1.04, both passing all physics checks --- the correct verdict is "inconclusive, propose discriminating experiment X."

- **CONDITIONALLY SUPPORTED:** A hypothesis that is valid *if* certain assumptions hold, where those assumptions are not yet verified. Example: "This Kondo effect explanation is supported IF the impurity concentration is below 0.1%, which has not been measured." The current categories force this into "PLAUSIBLE," losing the critical conditional information.

- **SUPERSEDED:** A hypothesis that was once valid but has been replaced by a more complete theory that the system found. Distinct from REJECTED (which implies the hypothesis is wrong) --- SUPERSEDED means it is a valid approximation in a limited regime.

**Suggestion:** Add at minimum INCONCLUSIVE and CONDITIONALLY_SUPPORTED. For CONDITIONALLY_SUPPORTED, require the reviewer to state the conditions explicitly.

---

### 6. `physics_domain` as a single string is an inadequate representation

**Severity: Warning**

`config.py` line 34:

> `physics_domain: str = "condensed_matter"`

This single string propagates to `get_checklist`, `subfield_defaults`, and `lookup_pattern`. The available domains listed in prompts (`condensed_matter, qft, gr, amo, plasma`) are coarse-grained and mutually exclusive.

Real examples where this fails:
- **Cavity QED:** Requires both AMO and condensed matter conventions (for the cavity/circuit part in circuit QED).
- **Cosmological perturbation theory:** Requires both GR and QFT conventions (metric signature, Fourier conventions for cosmological perturbations vs. flat-space QFT).
- **Neutron star physics:** Nuclear physics + GR + plasma physics, each with different unit conventions.

If the GPD tools return domain-specific checklists, a single domain means the reviewer will miss checks relevant to the other domains involved. For instance, a neutron star equation-of-state problem with `physics_domain="gr"` would get GR checks but miss nuclear physics constraints (saturation density, symmetry energy bounds) that are equally important.

**Suggestion:** Accept `physics_domains: list[str]` and query checklists/conventions for each. The convention merger logic should flag conflicts (e.g., metric signature $(+,-,-,-)$ in particle physics vs. $(-,+,+,+)$ in GR) and present them explicitly.

---

### 7. No mechanism for the system to discover it has the wrong domain

**Severity: Note**

The current flow is: user sets `physics_domain` in config -> conventions locked -> literature search -> fitting -> review. At no point does any agent have the ability to say "the initially specified domain is wrong or incomplete." If a literature agent discovers that the phenomenon is best explained by a hypothesis from a different domain, there is no feedback path to update the conventions or checklist domain.

**Suggestion:** After the literature phase identifies hypotheses, add a domain-validation step where the system checks whether the hypotheses' domains match `config.physics_domain`. If they diverge, prompt the user or automatically expand the domain set.

---

### 8. The example "chi-squared=1.5 all PASS > chi-squared=0.9 check 5.1 FAIL" is physically misleading

**Severity: Note**

From `debate.py` line 56:

> `A model with chi-squared=1.5 and all checks PASS should rank above chi-squared=0.9 with check 5.1 FAIL.`

While this specific example is defensible (a dimensionally inconsistent model is wrong regardless of fit quality), it sets a dangerous precedent. A reduced chi-squared of 0.9 with a "5.1 FAIL" is obviously bad --- the failure is catastrophic. But the rigid rule this example teaches the LLM could be overapplied: a model with reduced chi-squared = 5.0 and all checks PASS should NOT rank above a model with reduced chi-squared = 1.01 and a marginal 5.3 flag (e.g., a limiting case recovered to 5% accuracy in a regime where 5% corrections are expected from higher-order terms).

The system conflates *severity* of check failures with *binary pass/fail*. A model that "fails" a limiting case check because it has known O(epsilon^2) corrections in that limit is physically fine.

**Suggestion:** Introduce severity levels for check results (FAIL-critical, FAIL-minor, WARNING, PASS) and adjust the ranking logic accordingly. A FAIL-critical on 5.1 (wrong dimensions) should indeed override chi-squared; a FAIL-minor on 5.3 (limiting case off by expected correction terms) should not.

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | Critical | Rigid ranking hierarchy fails when all models pass the same checks or when parsimony misleads |
| 2 | Critical | First-principles bias systematically disadvantages effective theories that are the correct answer in condensed matter, nuclear, and astrophysics |
| 3 | Warning | Mandatory check set missing conservation laws (5.4) and convergence (5.5); symmetry check inappropriate for broken-symmetry phenomena |
| 4 | Warning | Convention locking before literature search creates domain mismatch risk for cross-domain phenomena |
| 5 | Warning | Verdict categories lack INCONCLUSIVE and CONDITIONALLY_SUPPORTED, leading to overconfident conclusions |
| 6 | Warning | Single `physics_domain` string cannot represent cross-domain phenomena |
| 7 | Note | No feedback mechanism to correct wrong initial domain assignment |
| 8 | Note | Example in debate prompt teaches binary pass/fail when check severity should be graded |

The core theme: the system assumes physics can be verified by a fixed hierarchy of binary checks, but experimental physics frequently involves effective theories, crossover regimes, broken symmetries, and multi-domain phenomena where this assumption breaks down. The GPD integration is a valuable addition, but the deduction logic needs more flexibility to avoid systematically wrong conclusions in these common scenarios.

FaroutYLq:
## Integration & Testing Review

Reviewed all 12 changed files against the existing test suite and CLI entry point. This review focuses on integration completeness, test coverage, backward compatibility, and edge cases. Architecture and physics correctness are out of scope (handled by other reviewers).

---

### 1. :red_circle: Critical — `enable_gpd_mcp` defaults to `True` but `mcp` is an optional extra

`MTFConfig.enable_gpd_mcp` defaults to `True` (config.py L33). This means every user who runs `pip install mtf` (without `[gpd]`) will:
- Spawn a background daemon thread in `GPDMCPClient.__init__()` (gpd_mcp.py L50-54)
- Hit a `ModuleNotFoundError` for `mcp` inside `_start_async()`, caught by the `start()` try/except
- Get a warning log but otherwise proceed

The thread is created unconditionally in `__init__`, even before `start()` is called. If `start()` fails, `close()` is still called in the `finally` block (orchestrator.py L120-121), which calls `self._loop.call_soon_threadsafe(self._loop.stop)` — this works, but it means every default run pays the cost of creating and tearing down a background event loop + thread for nothing.

**Fix:** Either (a) default `enable_gpd_mcp=False` so GPD is opt-in, or (b) defer thread creation to `start()` so `__init__` is cheap, or (c) guard the `GPDMCPClient()` constructor in the orchestrator with a pre-check like `importlib.util.find_spec("mcp")`.

---

### 2. :red_circle: Critical — CLI does not expose any GPD config fields

`cli.py` was not modified in this PR. The new config fields (`enable_gpd_mcp`, `physics_domain`, `gpd_servers`) have no CLI arguments. Users cannot:
- Disable GPD: `--no-gpd` or `--disable-gpd-mcp`
- Set their physics domain: `--physics-domain qft`
- Select specific GPD servers

The `MTFConfig` constructed in `cli.py` (L54-64) does not pass `enable_gpd_mcp`, `physics_domain`, or `gpd_servers`, so they silently get the dataclass defaults. Since `enable_gpd_mcp=True` by default, CLI users with GPD installed will always get GPD; CLI users without GPD will always get a warning logged with no way to suppress it.

**Fix:** Add at minimum `--physics-domain` and `--no-gpd` / `--enable-gpd` flags to `build_parser()`, and pass them through to `MTFConfig`.

---

### 3. :red_circle: Critical — No tests added for any new functionality

The PR adds ~490 lines of new/changed code across 12 files and zero new tests. Specific gaps:

**Missing test cases (minimum set):**

| # | Test | Why it matters |
|---|------|---------------|
| T1 | `GPDMCPClient` lifecycle: `start()` when `mcp` is not installed sets `available=False` | Verifies graceful degradation for the default install path |
| T2 | `GPDMCPClient.call()` returns `""` when server is not in `_sessions` | Verifies the None-session guard (gpd_mcp.py L122-124) |
| T3 | `GPDMCPClient.make_tool()` returns `None` when server not started | Verifies the filter pattern used in all three phases |
| T4 | `LiteratureAgent.__init__` with `gpd_tools=None` | Backward compatibility — no GPD tools, no crash |
| T5 | `LiteratureAgent.__init__` with `gpd_tools=[]` | Empty list should not change behavior |
| T6 | `FittingAgent.__init__` with `gpd_tools=None` | Same as T4 |
| T7 | `ReviewerAgent.__init__` with `gpd_tools=None` | Same as T4 |
| T8 | `DebateEngine.synthesize()` with `phase="fitting"` includes physics ranking criterion in system prompt | The phase-conditional logic (debate.py L43-57) is untested |
| T9 | `DebateEngine.synthesize()` with `phase="literature"` does NOT include physics ranking criterion | Negative case of T8 |
| T10 | `DebateEngine.synthesize()` injects `CONVENTIONS` and `PHYSICS_VERDICT` entries when present in memory | debate.py L69-74 is new and untested |
| T11 | `SharedMemory.filter(MemoryKind.CONVENTIONS)` and `filter(MemoryKind.PHYSICS_VERDICT)` | New enum members should have basic filter tests |
| T12 | Phase functions with `gpd=None` (backward compat) | Ensures all three phases work without GPD |
| T13 | `run_literature_phase` locks conventions into memory when `gpd` is provided | literature_phase.py L42-49 |

---

### 4. :yellow_circle: Warning — Existing `test_phases.py` tests still pass, but only by accident

The existing tests call `run_literature_phase()` without the `gpd` argument (test_phases.py L57-63, L93-99). This works because the new parameter has a default value (`gpd: Any | None = None`). However:

- The tests do not verify that `gpd=None` results in no GPD tool injection (the `gpd_tools` list should be empty).
- The `LiteratureAgent` mock in these tests (`MockAgent`) doesn't validate what `gpd_tools` kwarg it receives, so if someone accidentally changes the default to a non-None value, the tests won't catch it.

**Fix:** Add an assertion in existing tests that `MockAgent` was called with `gpd_tools=[]` (or at least that the kwarg is present and empty).

---

### 5. :yellow_circle: Warning — GPD server crash mid-run is partially handled but may leave zombie threads

If a GPD MCP subprocess crashes during a run:
- `call()` catches the exception and returns `""` (gpd_mcp.py L131) — good
- `make_tool()` returns a closure that calls `call()`, so it will also gracefully return `""` — good
- But the background event loop thread is a daemon thread that runs forever (`self._loop.run_forever`). If `close()` is not called (e.g., if the orchestrator itself crashes between `try` and `finally`), the daemon thread is orphaned. This is acceptable for a CLI tool but would be a problem in a library/notebook context.

**Fix (low priority):** Consider making `GPDMCPClient` a context manager (`__enter__`/`__exit__` and `__aenter__`/`__aexit__`) so it can be used with `with`/`async with` statements.

---

### 6. :yellow_circle: Warning — `_SERVER_MODULES` maps to `gpd.mcp.servers.*` but the PyPI package is `get-physics-done`

The module paths in `_SERVER_MODULES` (gpd_mcp.py L22-28) assume the GPD package installs under `gpd.mcp.servers.*`. This is an implicit coupling — if the GPD package restructures its internal module layout, this breaks silently (the server process would fail to start, caught by the try/except, logged as a warning, and the tool would be None).

**Fix:** Add a comment documenting which version of `get-physics-done` these module paths correspond to. Consider adding an integration test (marked `@pytest.mark.network` or similar) that verifies at least one server can start.

---

### 7. :yellow_circle: Warning — Unknown `physics_domain` silently produces empty/garbage conventions

If a user sets `physics_domain="biophysics"` (or any string GPD doesn't recognize), `gpd.call("conventions", "subfield_defaults", domain="biophysics")` will either return an empty string or an error message. The code (literature_phase.py L43-45) stores whatever comes back into `MemoryKind.CONVENTIONS` without validating it.

**Fix:** Check if the returned conventions string is non-empty and looks valid before storing it. At minimum, log a warning if the conventions response is empty.

---

### 8. :blue_circle: Note — `CLAUDE.md` and README not updated

Neither `CLAUDE.md` nor `README.md` were changed. The repository layout section in `CLAUDE.md` does not mention `gpd_mcp.py`. The Architecture Notes section does not describe GPD integration. The config fields table is missing `enable_gpd_mcp`, `physics_domain`, and `gpd_servers`. The Build & Development Commands section does not mention `pip install -e ".[gpd]"`.

---

### 9. :blue_circle: Note — Thread safety assumption in `GPDMCPClient` deserves a comment

`GPDMCPClient` uses a dedicated background thread with its own event loop. The `_sessions` dict is written on the background thread (inside `_start_async`) and read on the main thread (inside `call()` and `make_tool()`). This is safe only because `start()` blocks until `_start_async` completes (via `future.result()`), establishing a happens-before relationship. But this invariant is subtle and undocumented.

---

### 10. :blue_circle: Note — `sdk.Tool.from_function` may not accept functions without type annotations

`make_tool()` creates a closure `_tool_fn(**kwargs)` with no parameter type hints (gpd_mcp.py L145-148). Depending on how `sdk.Tool.from_function` introspects the callable (it likely uses `inspect.signature`), this may produce a tool with no input schema, which could cause the model to pass arguments incorrectly. This should be verified with a simple integration test.

---

### Summary

| Severity | Count |
|----------|-------|
| :red_circle: Critical | 3 |
| :yellow_circle: Warning | 4 |
| :blue_circle: Note | 3 |

The main concern is that this PR introduces a significant new subsystem (GPD MCP integration) with zero test coverage and incomplete CLI wiring. The graceful degradation path (GPD not installed) works by accident of exception handling rather than by design, and the default config (`enable_gpd_mcp=True`) means every user hits this path. I'd recommend at minimum:

1. Default `enable_gpd_mcp=False` (opt-in, not opt-out)
2. Add CLI flags for `--physics-domain` and `--enable-gpd`
3. Add the test cases listed in finding #3 (T1-T13)
4. Update `CLAUDE.md` repository layout and architecture sections

PR #5, opened 2026-03-21
Add Streamlit browser GUI (mtf-gui)

FaroutYLq:
## Summary

- **`mtf/gui.py`** — `StreamlitInterface` (implements `HumanInterface` ABC) + full Streamlit app
- **`pyproject.toml`** — `gui = ["streamlit>=1.35.0"]` optional dep group + `mtf-gui` script entry point
- **`README.md`** — new "Browser GUI" quick-start section with screenshot, updated install options
- **`CLAUDE.md`** — updated technology stack, repo layout, build commands, and Human Interface architecture note

## Interface screenshot

MTF Streamlit GUI — pre-run view

*Left sidebar: agent count sliders, debate rounds, physics domain multi-select, GPD toggle. Main area: phenomenon text box, drag-and-drop file uploader, Run Analysis button (enabled once text is entered).*

## Design: blocking-queue bridge

The orchestrator is a linear async coroutine; Streamlit reruns the entire script on every interaction. They are bridged with:

- **Background thread** — runs `MTFOrchestrator` in its own `asyncio` event loop; blocks on per-call `reply_q.get()` when awaiting user input
- **Streamlit thread** — drains `ui_queue` each rerun, renders messages, posts user responses back into the appropriate `reply_q`

Message protocol (`orchestrator → Streamlit`):

| Message | Payload | Reply |
|---|---|---|
| `"show"` | `content, title` | none (fire-and-forget) |
| `"ask"` | `prompt, reply_q` | `reply_q.put(str)` |
| `"confirm"` | `prompt, reply_q` | `reply_q.put(bool)` |
| `"done"` | `report` | — |
| `"error"` | `traceback` | — |

Each interactive message carries its own `reply_q` — no ordering ambiguity when phases interleave.

## Running view (approval gate)

When the orchestrator pauses for approval at the end of each debate round:

```
ℹ MTF asks for approval: Do you approve the literature synthesis?

   [ ✅ Approve ]     [ ✏️ Request Changes ]
```

Clicking either button calls `reply_q.put(bool)` to unblock the orchestrator thread, clears the pending widget, and triggers `st.rerun()`.

## Installation & usage

```bash
pip install -e ".[dev,gui]"
mtf-gui            # opens http://localhost:8501
```

## Test plan

- [ ] `pip install -e ".[dev,gui]"` installs without errors
- [ ] `mtf-gui` opens browser at `localhost:8501`
- [ ] Enter a phenomenon (no files), click Run — startup message appears
- [ ] Literature synthesis renders as collapsible expander, Approve button works
- [ ] Fitting and review phases render; final report shows with download button
- [ ] Upload a PNG — verify it appears in the image digest output
- [ ] Click "Request Changes" — verify feedback text box appears and next round starts
- [ ] Sidebar sliders affect `MTFConfig` values passed to the orchestrator
- [ ] Error during run — traceback rendered in `st.code` block


FaroutYLq:
## Code Review — PR #5: Add Streamlit browser GUI (`mtf-gui`)

Overall this is a clean, well-scoped addition. The blocking-queue bridge design is architecturally sound and the ABC contract is correctly implemented. Below is a detailed review.

---

### What is done well

- **Correct ABC implementation.** `StreamlitInterface` implements all three abstract methods (`show`, `ask`, `confirm`) and provides a sensible no-op override for `ask_for_files`, consistent with how files are wired in at startup. The docstring explaining _why_ `ask_for_files` returns `[]` is appreciated.
- **Per-message `reply_q` design.** Giving each interactive message its own `queue.Queue` instance (rather than a single shared reply queue) cleanly prevents ordering bugs if hypothetical future phases ever interleave `ask` and `confirm` calls.
- **Daemon thread.** Using `daemon=True` on the orchestrator thread means the process exits cleanly if the user closes the browser tab without finishing a run — no need for explicit join or shutdown logic.
- **`tmp_dir` lifetime management.** Storing the `tempfile.TemporaryDirectory` object in `ss.tmp_dir` keeps the directory alive for the duration of the session and lets Python's finalizer clean it up eventually (though see the leak note below).
- **Graceful GPD fallback.** Nothing in the GUI layer touches GPD directly; it all flows through `MTFConfig.enable_gpd_mcp`, which is the correct abstraction boundary.
- **`pyproject.toml` change is minimal.** Adding `gui` as an optional dep group and wiring `mtf-gui` as a script entry point is exactly the right pattern.

---

### Bugs and correctness issues

#### 1. `asyncio.get_event_loop()` deprecated / wrong API (`gui.py` lines 35, 41)

```python
return await asyncio.get_event_loop().run_in_executor(None, rq.get)
```

Inside `_run_orchestrator`, the background thread calls `asyncio.new_event_loop()` and `asyncio.set_event_loop(loop)` before running the orchestrator. By the time `StreamlitInterface.ask()` or `confirm()` is awaited, `asyncio.get_event_loop()` returns the correct loop on CPython 3.10+. However, `get_event_loop()` is deprecated in favour of `asyncio.get_running_loop()` (Python 3.10+). More importantly, if any future refactor removes the `set_event_loop` call (itself a deprecated pattern), this silently breaks. Replace with:

```python
return await asyncio.get_running_loop().run_in_executor(None, rq.get)
```

`get_running_loop()` raises `RuntimeError` if there is no running loop (making the bug loud instead of silent) and is the correct modern API inside a coroutine.

#### 2. `rq.get()` blocks indefinitely — no timeout (`gui.py` lines 35, 41)

If a Streamlit session is closed mid-run, the browser never sends a reply. `rq.get()` (called in the executor) blocks forever, keeping the orchestrator thread and its event loop alive. Since the thread is a daemon it dies when the _process_ exits, but a Streamlit server typically runs for hours, accumulating stuck threads. At minimum add `rq.get(timeout=300)` and propagate a `queue.Empty` as an `asyncio.CancelledError` or a descriptive exception, or add a `# TODO: add timeout` comment to document the known limitation.

#### 3. `tempfile.TemporaryDirectory` is never explicitly cleaned up (`gui.py` lines 150–152, 230)

`ss.tmp_dir` is removed from session state when the user clicks "Start new analysis", but `.cleanup()` is never called explicitly before the pop. `TemporaryDirectory.__del__` will eventually call it, but the timing is non-deterministic (CPython aside, the finalizer can be delayed arbitrarily). For a long-running server handling many sessions this leaks disk space. Add an explicit cleanup before the `ss.pop(...)` loop in the "Start new analysis" handler:

```python
if "tmp_dir" in ss:
    ss["tmp_dir"].cleanup()
```

#### 4. "Request Changes" UX — feedback text is collected in a separate subsequent interaction (`gui.py` lines 207–217)

When the user clicks "Request Changes", `reply_q.put(False)` is called and the orchestrator immediately unblocks and calls `ask()` to collect feedback text. That second call surfaces as a new `pending` widget with a blank text area — no context about what phase is asking. The flow is _correct_ but the UX is abrupt. Consider pre-rendering the feedback text area inside the `confirm` widget when "Request Changes" is clicked (one round-trip instead of two), and passing the combined `(False, feedback_text)` back. This is a UX issue rather than a correctness bug, but it will confuse users.

---

### Thread safety

The design is thread-safe by construction: the orchestrator thread _only_ writes to `ui_queue` and reads from individual `reply_q`s; the Streamlit thread _only_ reads from `ui_queue` and writes to `reply_q`s. `queue.Queue` is thread-safe. `st.session_state` is only accessed from the Streamlit thread. No shared mutable state without a lock. **This is correct.**

One minor point: `ss.messages` is rebuilt by concatenation on every `show` message:

```python
ss.messages = ss.get("messages", []) + [{"type": "show", ...}]
```

This allocates a new list on every call. For a long run with many `show` messages (each phase calls it multiple times) consider `ss.messages.append(...)` after ensuring the key is initialised as a list at startup — Streamlit session state handles mutable list objects correctly across reruns.

---

### Streamlit session state management

- **Mixed `.get()` vs attribute access.** The code mixes `ss.get("running")` (lines 128, 132, 166, 222) with direct attribute writes like `ss.running = True` (line 155). Both work in Streamlit session state — `.get()` returns `None` for missing keys, attribute access raises `AttributeError`. The pre-run guard on line 128 (`if not ss.get("running") and not ss.get("finished")`) correctly relies on `.get()` returning `None` on first load. Pick one convention and be consistent throughout; prefer `.get()` for reads before the key is guaranteed to exist, attribute access thereafter.

- **Polling interval.** The 300 ms `time.sleep(0.3)` + `st.rerun()` cycle (lines 219–222) is the standard polling pattern for Streamlit and is acceptable. Users will see up to ~300 ms latency between an agent posting a `show` message and it appearing on screen.

- **Session state not reset on error before "Start new analysis".** Correctly handled — `ss.running = False`, `ss.finished = True`, error is displayed, and the "Start new analysis" button clears everything.

---

### `main()` entry point pattern (`gui.py` lines 260–270)

```python
def main() -> None:
    import subprocess
    import sys
    subprocess.run(
        [sys.executable, "-m", "streamlit", "run", str(Path(__file__).resolve())],
        check=True,
    )
```

This is the standard and recommended pattern for wrapping a Streamlit app in a console-script entry point. `sys.executable` ensures the correct Python interpreter is used, and `Path(__file__).resolve()` correctly resolves to the installed file inside site-packages. **This is correct.**

One small issue: `check=True` raises `subprocess.CalledProcessError` when Streamlit exits with a non-zero code. The most common case is the user pressing Ctrl-C, which sends SIGINT to Streamlit and causes it to exit with code 130 (or 1), producing a noisy traceback. Consider:

```python
result = subprocess.run([...])
sys.exit(result.returncode)
```

---

### Minor style / consistency notes

1. **`asyncio.get_event_loop()` → `asyncio.get_running_loop()`** (lines 35, 41) — see bug #1 above; also a modernisation point.
2. **`files=files or None` comment** — line 83 passes `uploaded_paths or None`. `uploaded_paths` is `[]` when no files are uploaded and `[] or None` evaluates to `None`. This is correct but non-obvious; a one-line comment would help.
3. **Hardcoded `["condensed_matter"]` fallback** — line 103: `physics_domains=physics_domains or ["condensed_matter"]` duplicates the default in `MTFConfig`. If the default ever changes in `config.py`, this will silently diverge. A comment noting this intentional duplication (or referencing `MTFConfig` defaults) would help maintainability.
4. **Session state key list duplication** — the list of keys `["running", "finished", "ui_queue", "thread", "messages", "pending", "final_report", "error", "tmp_dir"]` appears both at initialisation time (lines 155–164) and in the cleanup loop (line 230). Define a module-level constant `_SESSION_KEYS` to keep them in sync.
5. **Deferred imports inside functions** — `streamlit`, `MTFConfig`, `MTFOrchestrator` are imported inside functions rather than at the top of the file. This is correct (avoids `ImportError` when `streamlit` is not installed) and consistent with the optional-dependency pattern, but means type checkers cannot see them. A `TYPE_CHECKING` block at the module level would be a cleaner pattern.

---

### Summary table

| Category | Status |
|---|---|
| Blocking-queue bridge design | Correct and clean |
| Thread safety | Sound |
| ABC contract fully implemented | Yes |
| `asyncio.get_event_loop()` deprecation | **Needs fix** — use `get_running_loop()` |
| Indefinite `rq.get()` block | **Needs timeout or documentation** |
| `TemporaryDirectory` leak | **Needs explicit `.cleanup()` call** |
| "Request Changes" UX | Minor — feedback text box appears in a separate step |
| `main()` entry point | Correct; minor Ctrl-C noise |
| Session state consistency | Minor — mixed `.get()` / attribute access |
| Message list allocation | Minor — consider `.append()` |

The two correctness issues most worth addressing before merge are the `get_event_loop()` → `get_running_loop()` switch and the explicit `tmp_dir.cleanup()` call. The indefinite `rq.get()` block is worth at least a `# TODO` comment or a bounded timeout.

---

*AI code review by Claude Sonnet 4.6 via Claude Code*

FaroutYLq:
All three issues from the review have been addressed in 33f0eda:

- **`get_event_loop()` → `get_running_loop()`** — `StreamlitInterface.ask()` and `confirm()` now use `asyncio.get_running_loop()`. Raises `RuntimeError` immediately if called outside a running loop rather than silently creating a new one.

- **Indefinite `rq.get()` block** — replaced with `_blocking_get()`, a helper that polls with a 1-second timeout and gives up after 1 hour (`_REPLY_TIMEOUT = 3600.0`). The orchestrator thread will unblock and exit cleanly if the browser tab is closed, ending the daemon-thread leak.

- **`TemporaryDirectory` cleanup** — `tmp_dir.cleanup()` is now called explicitly in three places: on `"done"`, on `"error"`, and when the user clicks "Start new analysis". Temp files are freed as soon as they're no longer needed.

PR #6, opened 2026-03-22
GPD deep integration: physics quality backbone throughout pipeline

FaroutYLq:
## Summary

- **Fixes critical gap**: `PHYSICS_VERDICT` was defined and wired into `DebateEngine.synthesize()` but nothing ever wrote to it — synthesis always ran against an empty verdict block. This PR fixes that and promotes GPD from a reviewer-phase overlay to an active physics quality backbone across all three phases.
- **8 additions** spanning auto domain classification, pre-exec convention checking, proactive fitting warnings, hypothesis plausibility screening, cross-phase pattern intelligence, skills server integration, and debate-level dimensional check postscripts.
- **23 new integration tests** with `MockGPD`; all existing tests pass unchanged with `gpd=None`.

## What changed

| Addition | Description |
|---|---|
| 1. Auto domain classification | `orchestrator._classify_domains()` calls `route_protocol` + `route_skill`, matches against 14 known domain names, overwrites `config.physics_domains` ephemerally for the run |
| 2. PHYSICS_VERDICT population | `_run_phase_physics_checks()` in fitting phase runs checks 5.1 + 5.3 per fit report **before** debate synthesis |
| 3. convention_check in fitting | `FittingAgent.fit()` calls `convention_check` post-code-gen via `asyncio.to_thread`, writes violations to `PHYSICS_VERDICT`, retries once |
| 4. Fitting warnings pre-fetch | `_prefetch_fitting_warnings()` fans out `lookup_pattern` + `check_error_classes` per hypothesis before fan-out; stored as `FITTING_WARNINGS` in agent context |
| 5. Plausibility screen | `_screen_hypothesis_plausibility()` runs `limiting_case_check` per candidate after literature debate, shows `[PASS]/[WARN]/[FAIL]` badges before approval gate; returns CRITICAL-FAIL set for optional auto-rejection |
| 6. Cross-phase pattern intelligence | Domain patterns pre-fetched at literature phase start (`DOMAIN_PATTERNS`); `add_pattern` exposed in both literature and fitting agent tool lists |
| 7. Skills server | `"skills": "gpd.mcp.servers.skills_server"` added to `_SERVER_MODULES`; `"skills"` in default `gpd_servers` |
[...]

PR #8, opened 2026-03-22
Optional fitting phase with qualitative evaluation fallback

FaroutYLq:
## Summary

- Adds `--no-fitting` CLI flag (and `fitting_enabled` config field) to skip quantitative fitting when users lack the necessary data
- When fitting is skipped, a new `QualitativeEvaluationAgent` fan-out runs instead, evaluating each hypothesis against theory, literature, and image-extracted data and producing SUPPORTED/PLAUSIBLE/SPECULATIVE/REJECTED verdicts
- `ReviewerAgent` reads `QUALITATIVE_EVAL` and `FITTING_SKIPPED` memory entries so the review phase adapts accordingly
- GUI: "Skip fitting phase" checkbox added to sidebar

## New files
- `mtf/agents/qualitative.py` — `QualitativeEvaluationAgent`
- `mtf/phases/qualitative_phase.py` — `run_qualitative_phase()`

## Docs updated
- Architecture overview: pipeline diagram with fitting/qualitative branch; no-fitting mode subsection
- `docs/architecture/agents.md`: `QualitativeEvaluationAgent` section
- `docs/architecture/memory.md`: `QUALITATIVE_EVAL`, `FITTING_SKIPPED` kinds; accumulation order showing both paths
- `docs/usage/cli.md`: `--no-fitting` flag

## Test plan
- [ ] `mtf "phenomenon" --no-fitting` runs without error and produces a qualitative evaluation + review report
- [ ] Without `--no-fitting`, existing fitting path is unchanged
- [ ] GUI checkbox correctly toggles `fitting_enabled`
- [ ] `ReviewerAgent` report mentions qualitative evaluation context when fitting was skipped

### DIFF
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -197,6 +197,29 @@ The fit output and hypothesis text are stored as `FIT_RESULT`.
 
 ---
 
+## QualitativeEvaluationAgent
+
+| | |
+|---|---|
+| **Phase** | ② Qualitative Evaluation (when `--no-fitting` is used) |
+| **API** | `sdk.query()` (agentic) |
+| **Tools** | GPD: `get_checklist`, `run_check`, `dimensional_check`, `limiting_case_check`, `check_error_classes`, `get_detection_strategy`, `lookup_pattern`, `add_pattern` |
+| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `USER_FEEDBACK`, `CONVENTIONS`, `PHYSICS_VERDICT`, `FITTING_WARNINGS` |
+| **Memory written** | `QUALITATIVE_EVAL` |
+
+Used instead of `FittingAgent` when the pipeline runs with `--no-fitting`. N instances run
+concurrently. Each evaluates all hypotheses against established theory, literature context,
+and image-extracted data — without numerical fitting. For each hypothesis, the agent produces:
+a theoretical plausibility assessment, expected observational signatures, the specific data
+that would be needed to upgrade the assessment to a quantitative fit, a verdict
+(SUPPORTED / PLAUSIBLE / SPECULATIVE / REJECTED), and the single most decisive measurement
+that would confirm or refute it. Results are synthesized via
+`DebateEngine.synthesize(phase="qualitative")` and stored as `QUALITATIVE_EVAL`. The
+`ReviewerAgent` reads `QUALITATIVE_EVAL` in its `extra_kinds` so the review phase adapts its
+tone accordingly.
+
+---
+
 ## ReviewerAgent
 
 | | |
@@ -204,7 +227,7 @@ The fit output and hypothesis text are stored as `FIT_RESULT`.
 | **Phase** | ③ Review |
 | **API** | `sdk.query()` (agentic) |
 | **Tools** | GPD: `get_checklist`, `run_check`, `dimensional_check`, `limiting_case_check`, `check_error_classes`, `get_detection_strategy`, `lookup_pattern`, `add_pattern` |
-| **Memory context read** | `LITERATURE`, `DEBATE`, `FIT_RESULT`, `USER_FEEDBACK`, `IMAGE_DATA`, `CONVENTIONS`, `PHYSICS_VERDICT` |
+| **Memory context read** | `LITERATURE`, `DEBATE`, `FIT_RESULT`, `USER_FEEDBACK`, `IMAGE_DATA`, `CONVENTIONS`, `PHYSICS_VERDICT`, `QUALITATIVE_EVAL`, `FITTING_SKIPPED` |
 | **Memory written** | `REVIEW` |
 
 K instances run concurrently.  `ReviewerAgent` reads the widest memory context of any
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -204,7 +204,7 @@ The fit output and hypothesis text are stored as `FIT_RESULT`.
 | **Phase** | ② Qualitative Evaluation (when `--no-fitting` is used) |
 | **API** | `sdk.query()` (agentic) |
 | **Tools** | GPD: `get_checklist`, `run_check`, `dimensional_check`, `limiting_case_check`, `check_error_classes`, `get_detection_strategy`, `lookup_pattern`, `add_pattern` |
-| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `USER_FEEDBACK`, `CONVENTIONS`, `PHYSICS_VERDICT`, `FITTING_WARNINGS` |
+| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `USER_FEEDBACK`, `CONVENTIONS`, `PHYSICS_VERDICT` |
 | **Memory written** | `QUALITATIVE_EVAL` |
 
 Used instead of `FittingAgent` when the pipeline runs with `--no-fitting`. N instances run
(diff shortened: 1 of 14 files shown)


FaroutYLq:
## PR Review: Optional Fitting Phase (#8)

### Summary

This PR adds a clean `--no-fitting` / `config.fitting_enabled = False` bypass that substitutes qualitative evaluation for numerical fitting. The overall structure is sound and the branching logic is correct end-to-end. There are no critical correctness bugs, but several important issues need attention before merge: the approval loop does not retry (unlike the fitting phase it replaces), `FITTING_SKIPPED` is written after synthesis rather than before (causing a race with the reviewer's memory context), and the qualitative phase reuses `config.n_fitting` for agent count without a dedicated config knob, which is confusing and will break if a user sets `n_fitting=0` thinking they are disabling fitting.

---

### Issues

#### 🔴 Critical

**None.** The branch executes correctly, memory kinds are registered, and the reviewer does read the new entries.

---

#### 🟡 Important

**1. No retry loop in the approval gate (`mtf/phases/qualitative_phase.py`, lines 103–108)**

The fitting phase approval gate is a proper loop: it re-runs the whole fan-out on rejection (bounded by `max_debate_rounds`). The qualitative phase runs exactly one round, then if the user rejects it, stores feedback and exits anyway — the synthesis that was already shown is what the reviewer will see. This breaks the invariant documented in the architecture overview ("rejected rounds store user feedback and repeat") and leaves the reviewer working from an unapproved synthesis.

Fix: wrap lines 75–108 in a `for round_idx in range(config.max_debate_rounds)` loop with `break` on approval, matching the pattern in `literature_phase.py`.

---

**2. `FITTING_SKIPPED` written after synthesis — reviewer may not see it in debate context (`mtf/phases/qualitative_phase.py`, lines 89–101)**

`FITTING_SKIPPED` is written at line 98, after `debate_engine.synthesize()` is called at line 89. `DebateEngine.synthesize()` calls `memory.format_context()` internally, which will not include `FITTING_SKIPPED` because it has not been written yet. More importantly, if the reviewer consumes `FITTING_SKIPPED` to adapt its tone but the synthesis text itself contains no mention of this flag, the reviewer and synthesis are misaligned.

The flag should be written **before** `debate_engine.synthesize()` is called, so that the synthesis receives it as context.

---

**3. `config.n_fitting` reused as qualitative agent count (`mtf/phases/qualitative_phase.py`, line 83; `mtf/config.py`)**

`QualitativeEvaluationAgent` instances are fanned out with `range(config.n_fitting)`. This is confusing because `n_fitting` semantically means "how many numerical fitting agents to spawn." When a user sets `--n-fitting 0` expecting to avoid spinning up fitting infrastructure, the qualitative phase will spawn zero agents and `asyncio.gather()` will return an empty list, causing `debate_engine.synthesize([])` to receive no reports — likely producing a meaningless synthesis rather than a clear error.

Also, the GUI slider labeled "Fitting agents" (`mtf/gui.py`, line 107) now silently controls the qualitative fan-out count when `skip_fitting=True`; the label is misleading.

Fix: add `n_qualitative: int = 3` to `MTFConfig`, use it in `run_qualitative_phase`, and update the GUI label or add a second slider.

---

**4. `ReviewerAgent` system prompt not adapted for the qualitative path (`mtf/agents/reviewer.py`, lines 19–24 and 40)**

The reviewer's system prompt instructs the agent to call `run_check` "for each fit result" and rank by chi-squared. When `FITTING_SKIPPED` is present and `FIT_RESULT` entries are absent, following this prompt literally will cause the reviewer to either hallucinate fit-check results or skip those steps silently without adapting its output structure. The prompt needs a conditional branch along the lines of: "If `FITTING_SKIPPED` is present in context, skip steps 3–4 (fit-result checks) and instead evaluate based on `QUALITATIVE_EVAL` verdicts. Do not reference chi-squared or numerical fit results."

---

**5. Dimensional-check postscript not triggered for `phase="qualitative"` — intentional or oversight? (`mtf/debate.py`, line 107)**

`_append_dimensional_check` only fires for `phase in ("fitting", "review")`. The qualitative synthesis frequently contains theoretical equations (the system prompt explicitly asks for "specific physical arguments"). If this omission is intentional (to keep the qualitative path lightweight), add a comment. If not, `"qualitative"` should be added to the set and the physics ranking criterion in `base_system` should also extend to `"qualitative"`.

---

**6. `FITTING_WARNINGS` in `extra_kinds` but never pre-fetched in the qualitative path (`mtf/agents/qualitative.py`, line 55; `mtf/phases/qualitative_phase.py`)**

`QualitativeEvaluationAgent.evaluate()` includes `MemoryKind.FITTING_WARNINGS` in `extra_kinds`. In the qualitative path, `_prefetch_fitting_warnings()` (which lives in `fitting_phase.py`) is never called, so this slot will always be empty. This is harmless but misleading — it implies data that is never written.

Options: (a) call `_prefetch_fitting_warnings()` at the start of `run_qualitative_phase` (pattern and error-class lookups are equally useful here), or (b) remove `FITTING_WARNINGS` from `extra_kinds` in `QualitativeEvaluationAgent`. Option (a) is strictly better for physics quality and is consistent with the GPD integration philosophy.

---

#### 🟢 Minor / Suggestions

**7. Tool description says "fit result" in a no-fitting context (`mtf/phases/qualitative_phase.py`, lines 44–48)**

The description for `run_check` passed to `make_tool` says "Run a specific physics verification check against **a fit result** or hypothesis text." In the qualitative path there are no fit results; changing "a fit result" to "hypothesis text or qualitative assessment" avoids confusing the model.

**8. `interface.show()` message says "no fitting data available" (`mtf/phases/qualitative_phase.py`, line 29)**

This implies fitting was attempted and failed. The more accurate framing is "fitting skipped by user request".

**9. Redundant list spread (`mtf/agents/qualitative.py`, line 34)**

`tools=[*extra_tools]` is equivalent to `tools=list(extra_tools)`. Since `extra_tools` is already a `list[Any]`, the spread adds no value.

**10. `memory.md` timeline ordering (`docs/architecture/memory.md`)**

The comment block shows `FITTING_SKIPPED` after `QUALITATIVE_EVAL`, but if fix #2 is applied (write the flag before synthesis), `FITTING_SKIPPED` should appear before `QUALITATIVE_EVAL` in the timeline to match actual write order.

**11. Unrelated linter fix in `fitting.py`**

The rename of `l` to `line` is a good cleanup but unrelated to this feature. Consider moving it to a separate commit to keep the diff focused.

---

### Missing Tests

The PR adds no tests. The following should exist:

1. **`test_qualitative_phase_writes_memory`** — run `run_qualitative_phase` with mocked agents and debate engine, assert `memory.filter(MemoryKind.QUALITATIVE_EVAL)` is non-empty and `memory.filter(MemoryKind.FITTING_SKIPPED)` has exactly one entry.

2. **`test_qualitative_phase_approval_loop`** — analogous to `test_literature_phase_loops_on_rejection`: confirm the phase loops on rejection (once fix #1 is applied), storing `USER_FEEDBACK` each time.

3. **`test_orchestrator_branches_on_fitting_disabled`** — patch both `run_fitting_phase` and `run_qualitative_phase`, set `config.fitting_enabled = False`, run the orchestrator, assert `run_fitting_phase` was not called and `run_qualitative_phase` was called once.

4. **`test_orchestrator_branches_on_fitting_enabled`** — inverse: `fitting_enabled=True` calls `run_fitting_phase`, not `run_qualitative_phase`.

5. **`test_reviewer_sees_qualitative_eval_memory`** — populate `SharedMemory` with `QUALITATIVE_EVAL` and `FITTING_SKIPPED` entries, call `ReviewerAgent._build_prompt()` (or equivalent helper), assert both kinds appear in the resulting context string.

6. **`test_new_memory_kinds_round_trip`** — add entries for `QUALITATIVE_EVAL` and `FITTING_SKIPPED` to `SharedMemory`, verify `filter()` and `format_context()` return them correctly (extends `test_memory.py`).

7. **`test_no_fitting_cli_flag`** — unit-test `build_parser()` with `["phenomenon", "--no-fitting"]`, assert `args.no_fitting is True` and that `not args.no_fitting` yields `fitting_enabled=False`.

---

### Docs

- **`docs/architecture/overview.md` Mermaid diagram**: the diagram shows the qualitative path but does not include a node for writing `FITTING_SKIPPED`, unlike how `FC` is shown writing `PHYSICS_VERDICT` in the fitting path. A `QS["Write FITTING_SKIPPED"]` node between `QD` and `QU` would make the memory write order visually clear.

- **`docs/architecture/overview.md`, "No-Fitting Mode" section**: states "The qualitative phase runs an approval loop (same as the fitting phase)." This is currently inaccurate — it runs only a single confirmation prompt. Remove or correct this sentence once fix #1 is applied.

- **`docs/architecture/agents.md`, QualitativeEvaluationAgent table**: "Memory context read" lists `FITTING_WARNINGS` but warnings are never pre-fetched in the qualitative path (issue #6). Either update the table or add the pre-fetch call.

- **`docs/usage/cli.md`**: the flag is documented correctly. No issues.

---

### Verdict

**REQUEST CHANGES** — primarily for issues #1 (no retry loop), #2 (`FITTING_SKIPPED` written too late), and #4 (reviewer system prompt not adapted for the qualitative path). Issues #3 and #6 are important correctness concerns at edge cases. The structural design is good and the diff is clean; all of these are fixable without redesign.

PR #9, opened 2026-03-22
Measurement proposal sub-agents inside review phase

FaroutYLq:
## Summary

- Adds `ProposalAgent` — dedicated sub-agents that run in parallel with `ReviewerAgent` inside the review phase and produce a prioritized list of new measurements
- Each proposal specifies: observable, expected signal per hypothesis, discriminating power (HIGH/MEDIUM/LOW), equipment requirements, and required sensitivity
- Proposals are synthesized separately (`DebateEngine.synthesize(phase="proposals")`), stored as `MemoryKind.PROPOSALS`, and appended to the final report as `## Proposed Measurements`
- GUI: "Proposal agents" slider added to sidebar

## New files
- `mtf/agents/proposal.py` — `ProposalAgent`

## Config additions
- `n_proposal: int = 2`
- `proposal_model: str = "claude-opus-4-6"`

## Docs updated
- Architecture overview: review phase diagram shows parallel reviewer + proposal fan-out
- `docs/architecture/agents.md`: `ProposalAgent` section
- `docs/architecture/memory.md`: `PROPOSALS` kind
- `docs/usage/cli.md`: `--n-proposal`, `--proposal-model` flags
- `docs/usage/gui.md`: proposal agents sidebar control

## Test plan
- [ ] Final report contains a `## Proposed Measurements` section
- [ ] Proposals are ordered HIGH → MEDIUM → LOW discriminating power
- [ ] `--n-proposal 1` reduces proposal agents to 1
- [ ] Works in both fitting and `--no-fitting` modes (when combined with that branch)

### DIFF
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -272,7 +272,7 @@ The review report is stored as `REVIEW`.
 | **Phase** | ③ Review (parallel with ReviewerAgent) |
 | **API** | `sdk.query()` (agentic) |
 | **Tools** | GPD: `lookup_pattern`, `check_error_classes` |
-| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `FIT_RESULT`, `USER_FEEDBACK`, `REVIEW`, `CONVENTIONS`, `PHYSICS_VERDICT` |
+| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `HYPOTHESIS`, `FIT_RESULT`, `USER_FEEDBACK`, `CONVENTIONS`, `PHYSICS_VERDICT` |
 | **Memory written** | `PROPOSALS` |
 
 Runs concurrently with `ReviewerAgent` instances inside the review phase. Proposes a prioritized set of new measurements and experiments that would best discriminate between competing hypotheses. Each proposal includes: observable to measure, expected signal per hypothesis, discriminating power (HIGH / MEDIUM / LOW), equipment requirements, and required sensitivity. A "Bottom line" recommendation names the single most cost-effective measurement. Results are synthesized via `DebateEngine.synthesize(phase="proposals")` into a deduplicated, ranked proposals list that is appended to the final report as `## Proposed Measurements`.
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -265,6 +265,20 @@ The review report is stored as `REVIEW`.
 
 ---
 
+## ProposalAgent
+
+| | |
+|---|---|
+| **Phase** | ③ Review (parallel with ReviewerAgent) |
+| **API** | `sdk.query()` (agentic) |
+| **Tools** | GPD: `lookup_pattern`, `check_error_classes` |
+| **Memory context read** | `IMAGE_DATA`, `LITERATURE`, `DEBATE`, `FIT_RESULT`, `USER_FEEDBACK`, `REVIEW`, `CONVENTIONS`, `PHYSICS_VERDICT` |
+| **Memory written** | `PROPOSALS` |
+
+Runs concurrently with `ReviewerAgent` instances inside the review phase. Proposes a prioritized set of new measurements and experiments that would best discriminate between competing hypotheses. Each proposal includes: observable to measure, expected signal per hypothesis, discriminating power (HIGH / MEDIUM / LOW), equipment requirements, and required sensitivity. A "Bottom line" recommendation names the single most cost-effective measurement. Results are synthesized via `DebateEngine.synthesize(phase="proposals")` into a deduplicated, ranked proposals list that is appended to the final report as `## Proposed Measurements`.
+
+---
+
 ## ToolBuilderAgent
 
 | | |
--- a/docs/architecture/memory.md
+++ b/docs/architecture/memory.md
@@ -37,6 +37,7 @@ class MemoryEntry:
 | `FITTING_WARNINGS` | Fitting phase (GPD) | Pre-dispatch pitfall warnings combining domain `lookup_pattern` results (sign-error, convergence-issue categories) and `check_error_classes` output per hypothesis.  Written by `_prefetch_fitting_warnings()` before the fitting fan-out.  Carries `domain` and `hypothesis` metadata. |
 | `DOMAIN_PATTERNS` | Literature phase (GPD) | Cross-session convention-pitfall patterns pre-fetched before the first literature fan-out via `lookup_pattern(category="convention-pitfall")`.  Carries `domain` and `source` metadata. |
 | `DOMAIN_CLASSIFICATION` | `MTFOrchestrator._classify_domains()` | Audit trail of auto-detected physics domains when `config.auto_detect_domains=True`.  Records either the detected list or a fallback notice.  Informational only — not consumed by agents via `extra_kinds`. |
+| `PROPOSALS` | `ProposalAgent` (via review phase) | Synthesized list of proposed new measurements, ordered by discriminating power. Written after the proposal debate synthesis. Appended to the final report. |
 | `TOOLKIT_DIGEST` | `ToolBuilderAgent` | Summary of data and model items parsed from complex user-supplied input by the tool-builder agent. |
 | `QUALITATIVE_EVAL` | `QualitativeEvaluationAgent` | Qualitative hypothesis evaluation report produced when `--no-fitting` is used. Contains per-hypothesis SUPPORTED/PLAUSIBLE/SPECULATIVE/REJECTED verdicts based on theory and image data, without numerical fitting. Read by `ReviewerAgent`. |
 | `FITTING_SKIPPED` | Qualitative phase | Flag entry written when the fitting phase is skipped. Content: `"Fitting phase was skipped (--no-fitting). Qualitative evaluation substituted."` Read by `ReviewerAgent` for context. |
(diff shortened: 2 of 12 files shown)


FaroutYLq:
## PR Review: Measurement Proposal Sub-Agents (#9)

### Summary

This PR adds `ProposalAgent` — a new agent type that runs concurrently with `ReviewerAgent` in the review phase and proposes prioritized experiments to discriminate between competing hypotheses. The overall design is sound: the gather-then-dual-synthesis pattern is correct, the config/CLI/GUI wiring is complete, and the docs are thorough. However, there is one critical semantic bug (agents reading memory written by their concurrent partners), one important design gap (proposals provide no value when `n_proposal=1`), and several smaller issues worth fixing before merge.

---

### Issues

#### 🔴 Critical

**`MemoryKind.REVIEW` in `extra_kinds` is a data hazard** (`mtf/agents/proposal.py`, line 60)

`ProposalAgent.propose()` lists `MemoryKind.REVIEW` in its `extra_kinds`. But reviewer agents and proposal agents are launched in the same `asyncio.gather()` call (`review_phase.py`, lines 107–110):

```python
all_results = await asyncio.gather(
    *(a.review(phenomenon) for a in agents),
    *(a.propose(phenomenon) for a in proposal_agents),
)
```

`ReviewerAgent.review()` writes `REVIEW` entries to `SharedMemory` at the end of its own `_query()` call — after the agentic sdk loop completes. Because `asyncio.gather()` runs all coroutines concurrently, and because `SharedMemory` has no synchronization, there is no guarantee (and in practice essentially zero chance) that any `REVIEW` entry from the *current* phase has been committed to memory before `ProposalAgent._build_prompt()` is called. The proposals will run against an empty `REVIEW` filter for the current run, silently falling back to whatever (if anything) was there from a *prior* run.

**Fix:** Either (a) run the reviewer fan-out first, await all results, then run the proposal fan-out, or (b) pass `reviewer_reports` (already a `list[str]`) directly to `ProposalAgent.propose()` as a `context_override` argument, or (c) remove `REVIEW` from `extra_kinds` entirely and rely on the rich `FIT_RESULT` + `DEBATE` + `PHYSICS_VERDICT` context that is already available. Option (c) is the simplest fix; option (a) has the cleanest semantics but adds latency.

---

#### 🟡 Important

**Proposal synthesis with `n_proposal=1` adds zero deduplication value** (`mtf/phases/review_phase.py`, lines 120–124; `mtf/debate.py`, lines 73–81)

The synthesis prompt for `phase="proposals"` is entirely about deduplicating and merging overlapping proposals from multiple agents. When `n_proposal=1` (which is a valid configuration, and the minimum enforced by the GUI slider), `DebateEngine.synthesize()` is called on a single-element list. The synthesis call is a full `messages.create()` round-trip with no value added — it just paraphrases the single report. This is API cost and latency with no benefit.

**Fix:** Short-circuit in `run_review_phase`: if `len(proposal_reports) == 1`, skip the synthesis call and use `proposal_reports[0]` directly (same pattern used elsewhere in the codebase for trivial cases).

**`review_synthesis` is never stored in `SharedMemory`** (`mtf/phases/review_phase.py`, line 114)

After the merge, `DebateEngine.synthesize(phase="review")` is called and returns `review_synthesis`. `DebateEngine.synthesize()` itself writes the result to `SharedMemory` as `MemoryKind.DEBATE` (see `debate.py`, line 119) — that part is fine. But the existing `MemoryKind.REVIEW` entries (the individual agent reports) were already written by each agent inside its own `_query()` call. The docs say `DEBATE` is the synthesis; that is already handled. No change needed here — but note that `proposal_synthesis` is also written to memory inside `review_phase.py` line 125 *in addition* to `DebateEngine.synthesize()` already having written it as `MemoryKind.DEBATE` at line 119. The result is that `proposal_synthesis` text is in memory **twice**: once as `DEBATE` (phase="proposals") and once as `PROPOSALS`. This is harmless but redundant. Consider either relying on the `DEBATE` entry alone or skipping `DebateEngine.synthesize()`'s internal `memory.add()` for the proposals phase.

**`proposal_model` is not forwarded from the GUI** (`mtf/gui.py`, lines 156–164)

`MTFConfig` is constructed in `_streamlit_app()` without `proposal_model`. The `n_proposal` slider is correctly wired, but the model is always the dataclass default (`"claude-opus-4-6"`). There is no GUI control for `proposal_model` and it is silently not passed to `MTFConfig`. This is a minor inconsistency but will be confusing if a user selects a different `reviewer_model` and expects it to apply to proposal agents too. Either add a slider for `proposal_model` or at minimum copy `reviewer_model` when constructing the config (since they run in the same phase).

**GPD tools not wired to P-agents in the Mermaid diagram** (`docs/architecture/overview.md`, line 60)

The existing `GPD -.->|"tools"| R` edge covers reviewer agents. Proposal agents also receive GPD tools (`lookup_pattern`, `check_error_classes`) but there is no edge from GPD to the `P` node. The diagram correctly shows `P` but leaves its tool source invisible.

---

#### 🟢 Minor / Suggestions

**Slice correctness is fine but brittle** (`review_phase.py`, lines 111–112)

```python
reviewer_reports = list(all_results[:config.n_reviewer])
proposal_reports = list(all_results[config.n_reviewer:])
```

This is correct given that reviewers are added to the gather first. It is fragile: if a future change reorders the generator expressions in `asyncio.gather()`, the slices silently produce wrong results with no runtime error. A more robust pattern is to gather them separately and `zip` or concatenate explicitly:

```python
reviewer_coros = [a.review(phenomenon) for a in agents]
proposal_coros = [a.propose(phenomenon) for a in proposal_agents]
all_results = await asyncio.gather(*reviewer_coros, *proposal_coros)
```

The current code does exactly this — the comment is just a readability note. Consider adding an assertion: `assert len(all_results) == config.n_reviewer + config.n_proposal` to catch ordering bugs early.

**`ProposalAgent` system prompt does not reference hypotheses explicitly** (`mtf/agents/proposal.py`, lines 10–26)

The prompt says "discriminate between the candidate hypotheses" but does not tell the agent *where* hypotheses are found. `HYPOTHESIS` entries are not in `extra_kinds`. The agent must infer hypotheses from `DEBATE` and `LITERATURE` context. This usually works, but adding `MemoryKind.HYPOTHESIS` to `extra_kinds` would make the context explicit and more reliable.

**`ProposalAgent.__init__` has a redundant list unpack** (`mtf/agents/proposal.py`, line 41)

```python
tools=[*extra_tools]
```

`extra_tools` is already a `list`. This should be `tools=extra_tools` (or `tools=list(extra_tools)` for defensiveness, but the star-unpack adds no value).

**`MemoryKind.PROPOSALS` comment is missing** (`mtf/memory.py`, line 25)

Every other `MemoryKind` value has an inline comment describing what it stores. `PROPOSALS = "proposals"` has none. Minor, but inconsistent.

**Double `interface.show()` for proposals** (`review_phase.py`, lines 127, 131)

`proposal_synthesis` is shown to the user at line 127 (`title="MTF: Proposed Measurements"`) and then the `final_report` (which embeds the same text verbatim under `## Proposed Measurements`) is shown again at line 131 (`title="Final Report"`). The user sees the proposals twice in the GUI's expanding panel list. The intermediate `show(proposal_synthesis)` call should likely be removed; showing the combined final report once is sufficient.

**`n_proposal` minimum in GUI is 1, not 0** (`mtf/gui.py`, line 109)

`st.slider("Proposal agents", 1, 4, 2)` — the minimum is 1. If a user wants to disable proposals entirely, they cannot do so from the GUI. Consider either making 0 a valid value (and skipping the whole proposal fan-out + synthesis when `n_proposal == 0`) or adding a checkbox to enable/disable. The CLI default of 2 will always launch proposals; there is no `--no-proposals` flag.

---

### Missing Tests

The PR adds no tests. Given the existing test conventions (`tests/test_phases.py` uses `patch` + `AsyncMock`), the following tests should exist:

1. **`test_review_phase_concurrent_gather_slicing`** — assert that with `n_reviewer=2` and `n_proposal=3`, `reviewer_reports` has length 2 and `proposal_reports` has length 3 after gather. Mock both `ReviewerAgent.review` and `ProposalAgent.propose`.

2. **`test_review_phase_writes_proposals_to_memory`** — assert that after `run_review_phase`, `memory.filter(MemoryKind.PROPOSALS)` has exactly one entry and its content matches the mocked proposal synthesis output.

3. **`test_review_phase_final_report_contains_proposals_section`** — assert that the return value of `run_review_phase` contains the string `"## Proposed Measurements"`.

4. **`test_proposal_agent_review_kind_not_in_memory_during_concurrent_run`** — a regression test that explicitly documents the hazard: run `propose()` concurrently with a stubbed `review()` that delays writing to memory, assert that `ProposalAgent` does not receive any `REVIEW` context from the current phase (this would fail today and become a guard rail once the bug is fixed).

5. **`test_debate_synthesize_proposals_phase`** — assert that when `phase="proposals"`, the system prompt sent to `messages.create` contains the deduplication instruction text (`"deduplicated"`, `"discriminating power"`), mirroring the existing `test_synthesize_stores_in_memory` pattern.

---

### Docs

- `docs/architecture/agents.md` — the `ProposalAgent` table says `Memory context read: REVIEW` without flagging that this is effectively a no-op during the same-phase concurrent run. The table should either remove `REVIEW` or add a note: "REVIEW from prior runs only — current-phase REVIEW entries are written concurrently and may not be available."

- `docs/architecture/memory.md` — the `PROPOSALS` row says "Written after the proposal debate synthesis." The `DEBATE` table row (for `phase="proposals"`) is not mentioned anywhere in the memory docs, but `DebateEngine.synthesize()` always writes to `DEBATE`. This means there will be a `DEBATE` entry with `phase="proposals"` that is not documented.

- `docs/architecture/overview.md` — the Mermaid diagram is otherwise accurate and the two-synthesis pattern (R→RD, P→PD, both→FR) correctly represents the code. The missing GPD arrow to `P` (noted above) is the only discrepancy.

- `docs/usage/cli.md` — the `--images` flag is documented but the actual CLI flag is `--files` (see `cli.py` line 34). This is a pre-existing discrepancy, not introduced by this PR, but worth flagging.

---

### Verdict

**REQUEST CHANGES** — the `MemoryKind.REVIEW` data hazard is a semantic correctness bug that will silently produce worse proposals in practice (agents miss all current-phase review context). The double `interface.show()` for proposals is a UX regression. The `proposal_model` omission in the GUI will surprise users. All three are straightforward fixes. The core design — concurrent gather, dual synthesis, correct slice, `PROPOSALS` memory kind, config/CLI wiring — is solid and worth building on.

PR #10, opened 2026-03-22
Enhanced PDF extraction with two-pass figure analysis

FaroutYLq:
## Summary

- Adds a two-pass extraction pipeline for dense PDFs (10+ pages, many figures) using no new dependencies
- **Pass 1** (general digest): existing `_PDF_SYSTEM_PROMPT`, now extended with a Figure Inventory section listing every figure by page number and caption
- **Pass 2** (figure extraction): new `_FIGURE_EXTRACTION_PROMPT` iterates page-by-page and extracts each figure individually — type, axes (labels/units/scale/range), all data series as numerical arrays, quantitative features, physical significance
- Enabled by default (`pdf_enhanced_extraction = True`); disable with `--no-enhanced-pdf`
- GUI: "Enhanced PDF extraction (2-pass)" checkbox added to sidebar

## Config additions
- `pdf_enhanced_extraction: bool = True`
- `pdf_figure_extraction_max_tokens: int = 4096`

## Docs updated
- `docs/usage/images.md`: major update — full PDF section with pass-by-pass description
- `docs/usage/cli.md`: `--no-enhanced-pdf` flag
- `docs/usage/gui.md`: enhanced PDF extraction sidebar control
- `docs/architecture/agents.md`: `FileDigestSubagent` updated for both passes
- `docs/architecture/overview.md`: Phase 0 updated with two-pass description

## Test plan
- [ ] A 20-page PDF with many figures produces a digest with both `## General Document Digest` and `## Figure-by-Figure Extraction` sections
- [ ] Each figure is individually enumerated with numerical data
- [ ] `--no-enhanced-pdf` produces only a single-pass digest (existing behaviour)
- [ ] Short PDFs (< 5 pages) still work correctly with enhanced mode on

### DIFF
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -51,10 +51,20 @@ A stateless, leaf-level worker that digests exactly one file.  It does not touch
   all axis labels and units and scale, every data series as a Python list of numbers,
   key quantitative features (peaks, plateaus, slopes, error bars, fit parameters),
   embedded annotations, and a brief physical interpretation.
-- **PDFs**: sends a content block of type `"document"` with base64 PDF data.  The system
-  prompt asks for: document type, title, authors, physical system and phenomena, key equations
-  (reproduced symbolically with symbol definitions), experimental methods and parameters,
-  all reported numerical values with units, and main conclusions.
+- **PDFs**: sends a content block of type `"document"` with base64 PDF data.
+
+  **Standard pass** (`_PDF_SYSTEM_PROMPT`): extracts document type, title, authors, physical
+  system, key equations (reproduced symbolically), experimental methods and parameters, all
+  reported numerical values with units, conclusions, and a **Figure Inventory** listing every
+  figure by page number and caption.
+
+  **Figure-extraction pass** (`_FIGURE_EXTRACTION_PROMPT`, enabled when
+  `config.pdf_enhanced_extraction = True`): a second API call with the same document block,
+  using a dedicated prompt that iterates page-by-page and extracts each figure individually —
+  type, axes (labels, units, scale, range), every data series as numerical arrays, key
+  quantitative features, and physical significance.  The two pass results are combined into a
+  single sectioned digest: `## General Document Digest` followed by
+  `## Figure-by-Figure Extraction`.
 
 The MIME type is detected via `mimetypes.guess_type()`; unrecognised formats default to
 `image/png`.
--- a/docs/architecture/overview.md
+++ b/docs/architecture/overview.md
@@ -99,9 +99,15 @@ can access extracted numerical data.
      type, axis labels and units, all data series as Python lists of numbers, key quantitative
      features (peaks, plateaus, slopes, error bars, fit parameters), embedded annotations, and
      a brief physical interpretation.
-   - **PDFs**: a separate system prompt asks for document type, physical system, key equations
-     (reproduced symbolically), experimental methods and parameters, all reported numerical
-     values with units and uncertainties, and conclusions.
+   - **PDFs**: processed in up to two passes when `config.pdf_enhanced_extraction = True` (default).
+     - *Pass 1 (general digest):* the full PDF is sent with `_PDF_SYSTEM_PROMPT`, which extracts
+       document metadata, physical system, key equations, experimental methods, all reported numerical
+       values, conclusions, and a Figure Inventory enumerating every figure by page.
+     - *Pass 2 (figure extraction):* the same PDF is sent again with `_FIGURE_EXTRACTION_PROMPT`,
+       which iterates page-by-page and extracts each figure individually — type, axes, data series
+       as numerical arrays, quantitative features, and physical significance.
+     - Both results are concatenated into a single structured digest.  When `pdf_enhanced_extraction = False`,
+       only Pass 1 runs (same as the pre-existing behaviour).
 3. Each digest is stored in `SharedMemory` as `MemoryKind.IMAGE_DATA` with
    `source_file` and `filename` metadata.
 4. If more than one file was provided, a second synthesis `messages.create()` call combines
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -27,6 +27,7 @@ mtf   # interactive mode — you are prompted for everything
 | `--no-fitting` | _(off)_ | Skip quantitative fitting; run qualitative hypothesis evaluation instead |
 | `--no-gpd` | _(off)_ | Disable GPD MCP physics verification |
 | `--gpd-servers` | _(all)_ | GPD servers to start |
+| `--no-enhanced-pdf` | _(off)_ | Use single-pass PDF extraction instead of the default two-pass figure analysis |
 
 ## Examples
 
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -2,7 +2,7 @@
 
 ```bash
 mtf "Describe your phenomenon here"
-mtf "Describe your phenomenon here" --images plot1.png plot2.png
+mtf "Describe your phenomenon here" --files plot1.png plot2.png
 mtf   # interactive mode — you are prompted for everything
 ```
 
@@ -22,7 +22,7 @@ mtf   # interactive mode — you are prompted for everything
 | `--proposal-model` | `claude-opus-4-6` | Model for proposal agents |
 | `--debate-model` | `claude-opus-4-6` | Model for debate synthesis |
 | `--image-digest-model` | `claude-opus-4-6` | Model for image digestion |
-| `--images` | _(none)_ | Image files to digest (PNG, JPG, GIF, WebP) |
+| `--files` | _(none)_ | Input files to digest (images: PNG, JPG, GIF, WebP; documents: PDF) |
 | `--physics-domains` | `condensed_matter` | GPD physics domains (space-separated) |
 | `--no-fitting` | _(off)_ | Skip quantitative fitting; run qualitative hypothesis evaluation instead |
 | `--no-gpd` | _(off)_ | Disable GPD MCP physics verification |
@@ -33,7 +33,7 @@ mtf   # interactive mode — you are prompted for everything
 
 ```bash
 # Quantum Hall phenomenon with two images
-mtf "Plateau in rho_xx at B=3T" --images rho.png Hall.png
+mtf "Plateau in rho_xx at B=3T" --files rho.png Hall.png
 
 # Cross-domain (neutron star)
 mtf "Neutron star cooling anomaly" --physics-domains gr nuclear amo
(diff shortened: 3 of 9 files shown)


FaroutYLq:
## PR Review: Enhanced PDF Extraction (#10)

### Summary

This PR adds a two-pass PDF extraction pipeline to `FileDigestSubagent`: Pass 1 runs the existing general-digest prompt (now augmented with a Figure Inventory section), and Pass 2 runs a new dedicated figure-extraction prompt. The feature is gated by `config.pdf_enhanced_extraction` (default `True`) and is wired cleanly through CLI (`--no-enhanced-pdf`) and GUI. The implementation is broadly sound but has several issues that should be addressed before merge — most critically around the token budget, the cost default being on without a page-count guard, and an undocumented variable-shadowing bug.

---

### Issues

#### 🔴 Critical

**1. `content_block` variable shadowed in the `else` branch — regression for images**
(`mtf/agents/image_digest.py`, lines 271–278)

In the original code, `content_block` was set once for whichever branch was taken (PDF or image) and then a single shared `response = await asyncio.to_thread(...)` call used it. The refactor moves the shared call inside both branches for PDFs, but the `else` branch that handles images now reassigns `content_block` to an image dict (lines 271–278) and calls `asyncio.to_thread()` correctly. This is fine _for images_ as a standalone run, but the assignment is preceded by the variable already being defined (as a PDF document block) in the outer `if media_type == "application/pdf":` scope. If Python ever reaches the `else` via a non-PDF file while `content_block` is already defined, the shadow is harmless — but the `system` variable **is never set** for images in the new code. Looking at line 279: `system = _IMAGE_SYSTEM_PROMPT` is inside the `else` block and that is fine. **However**, the `system` variable is not set at all in the PDF path — the calls there use the prompt names directly as keyword args, so there is no bug there. This is a false alarm on `system`, but confirms the refactor removed the unified response call correctly.

_Actual bug_: In the original diff the `else` branch for images is now indented under the outer `if media_type == "application/pdf":` block (note the diff context). Double-check that the `else` on line 270 correctly pairs with the `if media_type == "application/pdf":` on line 190 and not with the inner `if self._config.pdf_enhanced_extraction:` on line 205. If the indentation is wrong, image files would never be processed. The diff shows `else:` at the same indentation level as `if self._config.pdf_enhanced_extraction`, meaning it would pair with the _inner_ PDF-mode check — causing image files to only be reached when `pdf_enhanced_extraction` is `False`. **This is a structural bug**: when `pdf_enhanced_extraction=True`, image files would fall through without returning, raising `UnboundLocalError` on the implicit `None` return.

Concretely, the diff block is:
```python
            if self._config.pdf_enhanced_extraction:
                ...
                return (...)
            else:
                ...
                return response.content[0].text
        else:                     # <-- this should pair with the pdf check
            content_block = { ... image ... }
```
The indentation in the diff output shows the image `else:` is at column 8, same as `if self._config.pdf_enhanced_extraction:` at column 12 — meaning it pairs with the inner check, not with `if media_type == "application/pdf":`. **Verify in the actual file** that the image `else:` block aligns with `if media_type == "application/pdf":` (column 8). If not, this is a silent regression that breaks image processing whenever `pdf_enhanced_extraction=True` (the default).

**2. Pass 2 `max_tokens=4096` is almost certainly insufficient for dense PDFs**
(`mtf/config.py`, line 52; `mtf/agents/image_digest.py`, line 234)

A 20-page physics paper commonly contains 10–15 figures. Each figure entry in the structured extraction format (caption + type + axes + data arrays + key features + physical significance) is roughly 200–400 tokens per figure. 15 figures × 400 tokens = ~6,000 tokens minimum, before any preamble. With `max_tokens=4096` the model will be cut off mid-figure with no warning — `stop_reason` will be `"max_tokens"` rather than `"end_turn"`, but the code never checks `stop_reason`. The truncated text is silently concatenated into the digest and stored, and downstream agents will receive a structurally broken JSON-like block for the last few figures. 

**Fix required**: check `response2.stop_reason` and either (a) raise a warning, (b) log a note in the stored digest, or (c) increase the default `pdf_figure_extraction_max_tokens` to at least 8192. The current default of 4096 should not be the default for a feature marketed at "dense documents (many figures, 10+ pages)".

---

#### 🟡 Important

**3. No page-count guard — enhanced extraction is ON by default for all PDFs, including 1-pagers**
(`mtf/config.py`, `mtf/agents/image_digest.py`)

The docs say "For dense documents (many figures, 10+ pages)" but the code unconditionally runs two API calls for every PDF regardless of page count. A 1-page abstract or a 2-page supplement will incur double cost and double latency with no benefit. The Anthropic document API does expose `usage.input_tokens` in the response; alternatively, a simple `pdf_min_pages_for_enhanced` config field could gate the second pass. At minimum the docs should not frame this as a density-specific feature when it applies to all PDFs by default. Consider adding `pdf_min_pages_for_enhanced: int = 0` (disabled by default) as an opt-in threshold guard.

**4. `_synthesize()` `max_tokens` is fixed at 4096 and is not updated**
(`mtf/agents/image_digest.py`, line 384)

The synthesis call in `_synthesize()` is unchanged at `max_tokens=4096`. With enhanced extraction, individual digests sent to `_synthesize()` can be ~8k tokens each (Pass 1 + Pass 2 concatenated). For 3 files that is ~24k tokens of input. The synthesis output is then expected to produce a "Consolidated Quantitative Data" section merging all numerical data — 4096 output tokens will frequently be insufficient. This is a pre-existing limitation made significantly worse by this PR since individual digests are now much larger.

**5. The two `asyncio.to_thread()` calls are sequential, not concurrent**
(`mtf/agents/image_digest.py`, lines 207–252)

Both API calls for a single PDF are issued sequentially: `response1` must complete before `response2` starts. Since these are two independent calls to the Anthropic API with the same document block, they could be parallelised with `asyncio.gather()`. For a large PDF (5–10 second latency per call), sequential execution adds 5–10 seconds of dead wall-clock time per PDF. The `content_block` dict is a plain immutable dict and is safe to pass to two concurrent `asyncio.to_thread()` calls.

**6. `_FIGURE_EXTRACTION_PROMPT` does not elicit machine-readable arrays reliably**
(`mtf/agents/image_digest.py`, lines 99–122)

The prompt says `x = [...], y = [...]` as the target format, but this is shown as an example inside a natural-language template. Without an explicit instruction like "you MUST output numerical arrays as Python list literals, e.g. `x = [0.1, 0.5, 1.0]` — never use prose descriptions of values", the model will often substitute "approximately 0.1 to 1.0" or "see caption" for figures where digitisation is hard. The `_IMAGE_SYSTEM_PROMPT` (for regular images) includes the same instruction and has the same weakness, but for the figure-extraction pass the problem is amplified because the prompt structure (`**Data Series**: ...`) is parsed downstream by `FittingAgent`. A stricter instruction — e.g. "If values cannot be read precisely, output your best-estimate list and note the precision, but always output a list" — would reduce prose fallbacks.

**7. `--images` flag inconsistency in docs**
(`docs/usage/images.md`, lines 23–25 and 68; `docs/usage/cli.md`, line 22)

`docs/usage/images.md` instructs users to use `--images notes.pdf figure.png` but `cli.py` defines the flag as `--files` (line 36 of `cli.py`). The `--images` flag does not exist. The docs use `--images` throughout (lines 23, 68, 76). This is a pre-existing inconsistency but the PR's new docs repeat the wrong flag name. Anyone following the docs will get `error: unrecognized arguments: --images`.

---

#### 🟢 Minor / Suggestions

**8. `pdf_figure_extraction_max_tokens` is not exposed in the CLI or GUI**

The config field exists but there is no `--pdf-figure-extraction-max-tokens` CLI flag and no sidebar slider in the GUI. Advanced users who know they have a 30-figure paper cannot increase the token budget without writing Python. At minimum add a CLI flag.

**9. The `f`-string in `figure_user_text` is unnecessary**
(`mtf/agents/image_digest.py`, lines 225–230)

`figure_user_text` uses an `f`-string with `{path.name}` — fine — but `user_text` on lines 199–203 also uses an `f`-string only for `{path.name}`. Both are correct; this is just a note that `path.name` could be formatted once and reused.

**10. Hardcoded `max_tokens=4096` in Pass 1 is inconsistent with the config-driven Pass 2**
(`mtf/agents/image_digest.py`, line 210)

Pass 1 uses the literal `4096` while Pass 2 uses `self._config.pdf_figure_extraction_max_tokens`. There is no reason Pass 1 should not also use a config field (e.g. `pdf_general_digest_max_tokens`). As written, users cannot increase Pass 1's budget without editing source code.

**11. The docs describe `--images` not `--files`**
(see item 7 above — also affects `docs/usage/gui.md` where the interactive note says "When no `--images` flag is given" which is also wrong)

**12. `docs/usage/images.md` "Standard extraction" section is misleading**

Lines 27–38 describe "Standard extraction (single pass)" as a standalone section, but this section now describes what is _only_ Pass 1 of the default two-pass path. The heading implies this is what you get by default, which is the opposite of the truth. Rename to "Pass 1 — General digest" and move it under "Enhanced extraction" for clarity.

---

### Missing Tests

There are zero tests for `FileDigestSubagent` or `ImageDigestAgent`. This PR adds a significant branch (`pdf_enhanced_extraction`) with no test coverage. The following tests are needed:

1. **`test_file_digest_subagent_pdf_enhanced()`** — mock `anthropic.Anthropic().messages.create` to return two canned responses and assert:
   - Two calls are made (not one)
   - The returned string contains both `## General Document Digest` and `## Figure-by-Figure Extraction`
   - `response1` uses `_PDF_SYSTEM_PROMPT` and `response2` uses `_FIGURE_EXTRACTION_PROMPT`

2. **`test_file_digest_subagent_pdf_single_pass()`** — with `pdf_enhanced_extraction=False`, assert exactly one API call is made and the raw response text is returned unchanged.

3. **`test_file_digest_subagent_image_unaffected()`** — pass a `.png` file with `pdf_enhanced_extraction=True` and assert exactly one API call is made using `_IMAGE_SYSTEM_PROMPT`. This is the regression test for the indentation bug in item 1.

4. **`test_file_digest_subagent_stop_reason_truncation()`** — mock `response2.stop_reason = "max_tokens"` and assert a warning is logged or appended to the digest (once the fix for item 2 is implemented).

5. **`test_image_digest_agent_synthesis_with_enhanced_pdf()`** — assert that `_synthesize()` is called with the full concatenated digest (both passes) when multiple PDFs are provided.

---

### Docs

- `docs/usage/images.md` and `docs/usage/cli.md` both use `--images` which is not a valid CLI flag; the correct flag is `--files`. All examples need to be corrected.
- `docs/usage/images.md` line 43 says "enabled by default" which is accurate, but line 59 says "Useful when the PDF is short (< 5 pages)" — this is a reasonable heuristic but there is no automated enforcement of it in code. Either add the page-count guard (item 3) or soften the language to "consider disabling for short PDFs".
- `docs/architecture/agents.md` line 93 says "a third `messages.create()` call (the synthesis call)" — this should now say "a third or (for PDFs with enhanced extraction) fifth call" to be accurate about the total API call count.
- The Python API example in `docs/usage/images.md` lines 73–77 uses `orchestrator.run(..., images=[...])` — the actual signature uses `files=`, not `images=`. This is a pre-existing error the PR repeats.

---

### Verdict

**REQUEST CHANGES**

The indentation/structural bug (item 1) is a potential regression that breaks image processing in the default configuration and must be verified. The silent token truncation (item 2) will produce corrupted output for any dense PDF without any user-visible signal. The `--images` vs `--files` flag mismatch throughout the docs means the feature is not usable as documented. Fix these three issues and add the regression test for image files before merging.

PR #11, opened 2026-03-22
Deduplicate _strip_fences into shared mtf.utils.strip_fences

FaroutYLq:
## Summary

- `fitting.py` defined a private `_strip_fences()` function
- `tool_builder.py` had a byte-for-byte identical inline copy (11 lines) that was not reusing the existing function
- Extract both into `mtf/utils.py` as the single canonical `strip_fences()` and import it in both call sites

## Why

Using an existing function instead of duplicating logic is the minimal-reinvention approach. Any future bug fix or improvement only needs to happen in one place.

## Review notes (addressed)

- Trailing newline drop (`splitlines()` + `join()` silently loses a trailing `\n`) — pre-existing in both originals, no regression introduced
- Prose before the first fence is silently dropped — pre-existing, documented in new docstring
- Unclosed fence collects all remaining lines — pre-existing degraded behaviour, acceptable
- `_strip_fences` → `strip_fences` (drop leading underscore) is correct for a shared utility

## Test plan

- [x] `python -c "from mtf.utils import strip_fences"` succeeds
- [x] `from mtf.agents.fitting import FittingAgent` succeeds
- [x] `from mtf.agents.tool_builder import ToolBuilderAgent` succeeds
- [ ] Existing tests pass: `pytest tests/test_memory.py tests/test_debate.py tests/test_phases.py`

### DIFF
--- a/mtf/agents/fitting.py
+++ b/mtf/agents/fitting.py
@@ -9,6 +9,7 @@ from mtf.agents.base import BaseAgent
 from mtf.memory import MemoryKind, SharedMemory
 from mtf.toolkit.registry import ToolkitRegistry
 from mtf.tools.fitting_tools import run_fitting_code
+from mtf.utils import strip_fences
 
 if TYPE_CHECKING:
     from mtf.config import MTFConfig
@@ -42,22 +43,6 @@ experimental physics. Given a hypothesis and experimental data, you:
    runs on this domain.
 
 Always write clean, well-commented fitting code."""
-
-
-def _strip_fences(code: str) -> str:
-    """Strip markdown code fences from a code string."""
-    if "```" not in code:
-        return code
-    lines = code.splitlines()
-    code_lines = []
-    in_block = False
-    for line in lines:
-        if line.strip().startswith("```"):
-            in_block = not in_block
-            continue
-        if in_block:
-            code_lines.append(line)
-    return "\n".join(code_lines)
 
 
 class FittingAgent(BaseAgent):
@@ -131,7 +116,7 @@ class FittingAgent(BaseAgent):
                 MemoryKind.DOMAIN_PATTERNS,
             ),
         )
-        code = _strip_fences(code)
+        code = strip_fences(code)
 
         # Pre-exec convention check (Addition 3): check generated code for convention
         # violations before running it; retry once if violations are found.
@@ -176,7 +161,7 @@ class FittingAgent(BaseAgent):
                                 MemoryKind.DOMAIN_PATTERNS,
                             ),
                         )
-                        code = _strip_fences(code)
+                        code = strip_fences(code)
                 else:
                     break  # No violation — proceed to exec
 
--- a/mtf/agents/tool_builder.py
+++ b/mtf/agents/tool_builder.py
@@ -18,6 +18,7 @@ from typing import Any, Callable
 
 from mtf.agents.base import BaseAgent
 from mtf.memory import MemoryKind, SharedMemory
+from mtf.utils import strip_fences
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


FaroutYLq (commented):
## Code Review

This is a clean, well-scoped refactor. The duplication is real and the fix is the right call. A few observations, from most to least important:

---

### Behavioural equivalence between the two originals

The two originals are functionally identical, but they differed in one cosmetic detail:

- `fitting.py` had `code_lines = []` (untyped list)
- `tool_builder.py` had `code_lines: list[str] = []` (annotated)

The new `utils.py` picks up the annotation from `tool_builder.py`. That is the correct choice; no behaviour difference, just style.

**The call-site in `tool_builder.py` also changed semantics slightly.** The original was guarded by `if "` + '```' + `" in code:`, so if no fences were present the assignment was skipped entirely. The new code unconditionally calls `strip_fences(code)`, but because `strip_fences` itself opens with `if "` + '```' + `" not in code: return code`, the result is identical. The guard is now inside the function rather than at the call site. This is fine — callers no longer need to know the implementation detail.

---

### Edge case: trailing newline

`str.splitlines()` strips the trailing newline, so `"\n".join(code_lines)` will not restore it. A code string that originally ended with `\n` will lose that newline after going through `strip_fences`. Both originals had the same behaviour, so this PR introduces no regression. However it is worth noting as a latent inconsistency:

```python
# A trailing newline could be preserved with:
return "\n".join(code_lines) + ("\n" if code.endswith("\n") else "")
```

Whether this matters in practice depends on how `exec()` and `run_fitting_code()` handle the result — almost certainly not at all, but worth a conscious decision rather than an accident.

---

### Edge case: content before the first fence

If a model response includes prose before the opening fence (e.g. `"Here is the code:\n` + '```' + `python\nx = 1\n` + '```' + `"`), the implementation silently drops the prose — lines outside any fence block are ignored because `in_block` is `False`. That is the intended behaviour, and the updated docstring ("returns only the content that was inside the fences") documents it accurately. Both originals had identical behaviour here.

---

### Edge case: unclosed fence

If the model emits an opening fence with no closing fence, `in_block` ends as `True` and all lines from the opening to EOF are collected. The early-exit guard does not protect against this because the opening fence is present. The result — collecting the lines — is probably the right degraded behaviour rather than returning nothing. Both originals handled this identically.

---

### `re.sub()` alternative

A regex-based approach would be more compact and handle multiple blocks and edge cases in a single expression:

```python
import re

_FENCE_RE = re.compile(r"^```[^\n]*\n(.*?)^```[ \t]*$", re.MULTILINE | re.DOTALL)

def strip_fences(code: str) -> str:
    if "```" not in code:
        return code
    parts = _FENCE_RE.findall(code)
    return "".join(parts) if parts else code
```

That said, the line-scanning approach is readable and matches the style of the rest of the codebase. This is a suggestion, not a requirement.

---

### Module placement (`mtf/utils.py`)

Appropriate. The repo layout in CLAUDE.md shows no existing `utils.py`; putting a small pure-Python helper there is idiomatic. The module is not exported from `mtf/__init__.py`, which is correct — it is an internal implementation detail. If the utils module grows, a `__all__` list would help signal the intended public surface, but that is optional at this size.

---

### Naming: `_strip_fences` → `strip_fences`

Dropping the leading underscore is the right call for a shared utility. A private helper inside a single module does not need to be discoverable from outside; a shared utility should be.

---

### Summary

The extracted function is correct and equivalent to both originals. The module placement is appropriate. The only actionable suggestions are: (1) optionally preserve the trailing newline if that ever matters downstream, and (2) consider a regex implementation for compactness. Everything else is informational. **LGTM with no blocking issues.**

PR #12, opened 2026-03-22
Replace _looks_complex string triggers with stdlib compile()

FaroutYLq:
## Summary

- **Before:** `_looks_complex()` used a hardcoded tuple of trigger strings (`"\n"`, `"def "`, `"import "`, `"->"`, `"lambda "`, `"csv"`, `"\t"`) to decide whether user input needs the ToolBuilderAgent slow path vs a plain `eval()`
- **After:** Uses `compile(value, '<string>', 'eval')` — if it raises `SyntaxError`, the input is complex; otherwise it is a valid expression and takes the fast path

## Why this is better

| | Old approach | New approach |
|---|---|---|
| Accuracy | Heuristic — `[1,\n2,\n3]` falsely triggers the slow path | Exact — the Python compiler decides |
| Maintenance | Requires manually adding triggers as Python syntax evolves | Automatically correct for all syntax |
| Dependencies | None | None (stdlib `compile()`) |

## Behavior change: lambda expressions

`lambda x: x**2` previously matched the `"lambda "` trigger and went to the slow path (ToolBuilderAgent → `register_model`). It is now a valid `eval`-mode expression and goes to the fast path (`eval()` → `register_data`).

**This is accepted as intentional.** A lambda stored via `register_data` is still a callable the fitting agent can use. The correct long-term fix for the data/model bucket distinction is explicit user intent (a prompt asking whether the input is data or a model), not heuristic string matching.

## Review notes (addressed)

- `compile("", ..., "eval")` raises `SyntaxError` → `_looks_complex("")` returns `True`. Harmless in practice (caller already skips empty input)
- Single-line CSV-ish strings (e.g. `"1.0, 2.0, 3.0"`) now fast-path to a tuple; multi-line CSV still raises `SyntaxError` → slow path. Acceptable.
- The compiled code object is discarded rather than passed to `eval()` — minor inefficiency, kept simple for readability

## Test plan

- [x] `[1,2,3]`, `3.14`, `{'a': 1}` → fast path
- [x] `def f(): pass`, `import os`, `x = 1\ny = 2` → slow path
- [x] `lambda x: x` → fast path (intentional, see above)
- [ ] `pytest tests/test_phases.py`

### DIFF
--- a/mtf/phases/fitting_phase.py
+++ b/mtf/phases/fitting_phase.py
@@ -15,9 +15,18 @@ from mtf.toolkit.registry import ToolkitRegistry
 
 
 def _looks_complex(value: str) -> bool:
-    """Return True when the input is too rich for a plain eval()."""
-    triggers = ("\n", "def ", "class ", "import ", "->", "lambda ", "csv", "\t")
-    return any(t in value for t in triggers)
+    """Return True when the input is too rich for a plain eval().
+
+    Uses the stdlib compiler to check whether *value* is a valid single
+    expression.  If ``compile()`` with mode ``'eval'`` raises a
+    ``SyntaxError`` the string cannot be safely evaluated and the
+    ToolBuilderAgent slow path is needed.
+    """
+    try:
+        compile(value, "<string>", "eval")
+        return False
+    except SyntaxError:
+        return True
 
 
 async def _prefetch_fitting_warnings(


FaroutYLq (commented):
## Code Review

The direction is correct — replacing a brittle trigger-string heuristic with the actual Python compiler is strictly more accurate. But there are several edge cases worth working through before merging.

---

### 1. `compile(..., 'eval')` rejects valid "simple" inputs that the old code accepted

`'eval'` mode only accepts a **single expression**. Any syntactically valid Python that is a statement rather than an expression will now raise `SyntaxError` and be shunted to the slow path, even when the old heuristic would have let it through:

| Input | Old `_looks_complex` | New `_looks_complex` | Correct path? |
|---|---|---|---|
| `"lambda x: x**2"` | **True** (contains `"lambda "`) | False (lambda is an expression) | Fast path is fine here, but old code was over-eager |
| `"[1,\n2,\n3]"` | **True** (contains `"\n"`) | False (valid eval-mode expr) | Fast path is now correct |
| `"x = 1"` | False | **True** (assignment is a statement) | Slow path is correct |
| `""` (empty string) | False | **True** (SyntaxError on empty string) | Slow path for empty input seems wrong; the caller already guards `value.strip()` so this shouldn't matter in practice, but the function contract is surprising |

The biggest behaviour **regression** is `lambda x: x**2`. The old code sent it to the slow path (ToolBuilderAgent). The new code sends it to the fast path where `eval(value)` produces a callable. Then `toolkit.register_data(item_name, eval(value))` stores a function object as a "data" entry rather than a "model" entry. Whether that works depends on how downstream fitting code consumes the registry — it may be silently wrong. Consider whether `lambda` expressions should stay on the slow path or whether `register_model` should be used instead when the eval result is callable.

---

### 2. Multi-line string literals — behaviour improvement is real but worth documenting

```python
value = "[1, 2,\n3]"        # old: slow path (contains "\n"); new: fast path ✓
value = "{\n  'a': 1\n}"    # old: slow path; new: fast path ✓
```

These are genuine improvements. The old heuristic was too conservative for multi-line literals. The new code handles them correctly because `compile` parses the full grammar.

---

### 3. `"csv"` trigger is silently dropped

The old heuristic had `"csv"` as a trigger. A user pasting a CSV string like `"1.0, 2.0, 3.0\ntime,voltage\n..."` would previously route to the slow path. Now:

- If the CSV has a newline it still goes slow (SyntaxError — not a valid expression).
- If it is a single-line comma-separated value like `"1.0, 2.0, 3.0"` it goes to the fast path, `eval` succeeds producing a tuple `(1.0, 2.0, 3.0)`, and it gets registered as data. That is probably fine, but the intent of the old `"csv"` trigger was to catch the case where users paste tabular headers — make sure that still lands on the slow path (it does, because header lines aren't valid Python expressions).

---

### 4. Security: `compile()` is safe here; no concern

`compile()` with `mode='eval'` is a pure parse/bytecode step — it does **not** execute anything. There is no security regression compared to the old code. The actual `eval(value)` call that follows (line 196, already marked `# noqa: S307`) is unchanged and is the real execution point. Using `compile()` as a pre-check is the idiomatic pattern.

One small optimisation: the compiled code object returned by `compile()` is currently discarded. You could pass it directly to `eval()` instead of re-parsing, avoiding a second compile step. The calling site would look like:

```python
if not _looks_complex(value):
    try:
        toolkit.register_data(item_name, eval(compile(value, "<string>", "eval")))  # noqa: S307
        continue
    except Exception:
        pass
```

Or refactor `_looks_complex` to return the code object on success and `None` on failure. Minor optimisation, not a correctness issue.

---

### 5. Empty string edge case

`compile("", "<string>", "eval")` raises `SyntaxError`, so `_looks_complex("")` returns `True`. The caller already checks `if not value.strip(): continue` before calling `_looks_complex`, so this is harmless in practice. Still, it would be good to note in the docstring that the empty-string case is handled upstream, or add a short-circuit `if not value.strip(): return False` defensively to make the function self-contained.

---

### 6. Docstring is good but the `lambda` behaviour change deserves a mention

The new docstring explains the mechanism clearly. It would help to add one sentence noting that `lambda` expressions are now correctly identified as expressions (fast path) rather than being caught by the old `"lambda "` trigger (slow path), since this changes how lambda-provided callables are registered.

---

### 7. No new tests

The PR deletes a heuristic with several explicit triggers and replaces it with a fundamentally different mechanism. A few parametrised unit tests for `_looks_complex` would prevent regressions and make the intended behaviour explicit:

```python
@pytest.mark.parametrize("value, expected", [
    ("3.14", False),
    ("[1, 2, 3]", False),
    ("[1,\n2,\n3]", False),           # multi-line literal — was True before
    ("lambda x: x", False),           # lambda expr — was True before (behaviour change!)
    ("def f(x): return x", True),     # statement
    ("import numpy", True),           # statement
    ("x = 1", True),                  # assignment
    ("{'a': 1}", False),
    ("", True),                        # empty — slow path, handled upstream
])
def test_looks_complex(value, expected):
    assert _looks_complex(value) == expected
```

---

### Summary

The change is correct in principle and eliminates several false-positives (multi-line literals, valid expressions). The main concern is the **`lambda` behaviour change**: inputs like `lambda x: x**2` now go to the fast path and are registered via `toolkit.register_data()` instead of being handed to `ToolBuilderAgent` which would call `register_model`. Please verify this is intentional and won't silently break downstream fitting code that distinguishes data from model callables. Adding parametrised tests for the changed cases would also make the intent explicit and prevent future regressions.

FaroutYLq:
**Design decision re: lambda behavior change**

The reviewer correctly identified that `lambda x: ...` expressions now route to the fast path (`register_data`) instead of the slow path (ToolBuilderAgent -> `register_model`).

We are accepting this as intentional: a lambda stored via `register_data` is still a callable the fitting agent can use. The ToolBuilderAgent wrapping it in a named `def` is marginally better, but spawning a full agent for a one-liner is disproportionate. The correct long-term fix for the data/model bucket confusion is explicit user intent (a prompt asking whether the input is data or a model), not heuristic string matching. No code change needed.

FaroutYLq (commented):
## Code Review

The direction is correct — replacing a brittle trigger-string heuristic with the actual Python compiler is strictly more accurate. But there are several edge cases worth working through before merging.

---

### 1. `compile(..., 'eval')` rejects valid "simple" inputs that the old code accepted

`'eval'` mode only accepts a **single expression**. Any syntactically valid Python that is a statement rather than an expression will now raise `SyntaxError` and be shunted to the slow path, even when the old heuristic would have let it through:

| Input | Old `_looks_complex` | New `_looks_complex` | Correct path? |
|---|---|---|---|
| `"lambda x: x**2"` | **True** (contains `"lambda "`) | False (`lambda` is an expression) | Fast path is fine here, but old code was over-eager |
| `"[1,\n2,\n3]"` | **True** (contains `"\n"`) | False (valid `eval`-mode expr) | Fast path is now correct |
| `"x = 1"` | False | **True** (assignment is a statement) | Slow path is correct |
| `""` (empty string) | False | **True** (`SyntaxError` on empty string) | Slow path for empty input seems wrong; the caller already guards `value.strip()` so this shouldn't matter in practice, but the function contract is surprising |

The biggest behaviour **regression** is `lambda x: x**2`. The old code sent it to the slow path (ToolBuilderAgent). The new code sends it to the fast path where `eval(value)` produces a callable. Then `toolkit.register_data(item_name, eval(value))` stores a function object as a "data" entry rather than a "model" entry. Whether that works depends on how downstream fitting code consumes the registry — it may be silently wrong. Consider whether `lambda` expressions should stay on the slow path or whether `register_model` should be used instead.

---

### 2. Multi-line string literals — behaviour improvement is real but worth documenting

```python
value = "[1, 2,\n3]"   # old: slow path (contains "\n"); new: fast path ✓
value = "{\n  'a': 1\n}"  # old: slow path; new: fast path ✓
```

These are genuine improvements. The old heuristic was too conservative for multi-line literals. The new code handles them correctly because `compile` parses the full grammar.

---

### 3. `"csv"` trigger is silently dropped

The old heuristic had `"csv"` as a trigger. A user pasting a CSV string like `"1.0, 2.0, 3.0\ntime,voltage\n..."` would previously route to the slow path. Now:

- If the CSV has a newline it still goes slow (SyntaxError — not a valid expression).
- If it is a single-line comma-separated value like `"1.0, 2.0, 3.0"` it goes to the fast path, `eval` succeeds producing a tuple `(1.0, 2.0, 3.0)`, and it gets registered as data. That is probably fine, but the intent of the old `"csv"` trigger was to catch the case where users paste tabular headers — make sure that still lands on the slow path (it does, because header lines aren't valid Python expressions.

---

### 4. Security:  is safe here; no concern

 with  is a pure parse/bytecode step — it does **not** execute anything. There is no security regression compared to the old code. The actual  call that follows (line 196, already marked ) is unchanged and is the real execution point. Using  as a pre-check is the idiomatic pattern.

One small note: the compiled  object returned by  is discarded. You could pass it directly to  instead of re-parsing the string, which avoids a second parse:



And at the call site:



This is a minor optimisation, not a correctness issue.

---

### 5. Empty string edge case



So  returns . The caller already checks  before calling , so this is harmless in practice, but it would be good to note in the docstring that the empty-string case is handled upstream, or add  defensively.

---

### 6. Docstring is good but could mention the  behaviour change

The new docstring explains the mechanism clearly. It would help to add one sentence noting that  expressions are now correctly identified as expressions (fast path) rather than being caught by the old  trigger (slow path). This affects how -provided callables are registered.

---

### 7. No new tests

The PR deletes a heuristic with several explicit triggers and replaces it with a fundamentally different mechanism. A few parametrised unit tests for  would prevent regressions:



---

### Summary

The change is correct in principle and eliminates several false-positives (multi-line literals, valid expressions). The main concern is the ** behaviour change**: inputs like  now go to the fast path and are registered as data via  instead of being handed to  which would call . Please verify this is intentional and won't silently break downstream fitting code. Adding tests for the changed cases would also make the intent explicit and prevent future regressions.
EOF
)

PR #13, opened 2026-03-22
Replace _blocking_get polling loop with queue.Queue.get(timeout=...)

FaroutYLq:
## Summary

- Removes a manual poll-in-a-loop in `_blocking_get()` that reimplemented what `queue.Queue.get(timeout=N)` already provides
- Restores `import time` (needed by `time.sleep(0.3)` in `_streamlit_app()` — bug caught in review)

## Detail

The old implementation polled `rq.get(timeout=1.0)` in a `while` loop up to `_REPLY_TIMEOUT` (3600 s), advancing a monotonic deadline on each iteration. This is exactly what `queue.Queue.get(timeout=3600)` does internally via `threading.Condition` with an absolute deadline.

The docstring justified the polling as preventing daemon-thread GC leaks. Python daemon threads are killed automatically when the main process exits — no polling checkpoint needed. The simplified version is functionally equivalent with fewer moving parts.

## Review notes (addressed)

- **Critical bug fixed:** initial commit accidentally removed `import time`, but `time.sleep(0.3)` at line 282 still uses it → `NameError` on every GUI run. Fixed in follow-up commit.
- Signal handling unaffected — signals go only to the main thread regardless
- GIL release unaffected — `condition.wait()` inside `queue.Queue.get()` releases the GIL identically to the old approach
- Timeout precision is actually slightly better: OS timer granularity vs. up to 1 s late with the old 1-second poll interval

## Test plan

- [ ] Start the Streamlit GUI, submit a phenomenon, verify the UI responds to user input normally
- [ ] Closing the browser tab eventually lets the background thread exit (within 1 h)
- [ ] `pytest tests/` passes

### DIFF
--- a/mtf/gui.py
+++ b/mtf/gui.py
@@ -6,6 +6,7 @@ import asyncio
 import queue
 import tempfile
 import threading
+import time
 import traceback
 from pathlib import Path
 from typing import Any
--- a/mtf/gui.py
+++ b/mtf/gui.py
@@ -6,7 +6,6 @@ import asyncio
 import queue
 import tempfile
 import threading
-import time
 import traceback
 from pathlib import Path
 from typing import Any
@@ -17,20 +16,20 @@ from typing import Any
 
 from mtf.interface import HumanInterface
 
-_REPLY_POLL_INTERVAL = 1.0  # seconds between timeout checks in _blocking_get
-_REPLY_TIMEOUT = 3600.0     # 1 hour — abandon wait if browser tab was closed
+_REPLY_TIMEOUT = 3600.0  # 1 hour — abandon wait if browser tab was closed
 
 
 def _blocking_get(rq: "queue.Queue[Any]") -> Any:
-    """Poll *rq* in short intervals so the thread can eventually be GC'd if the
-    browser session disappears (avoids an indefinite daemon-thread leak)."""
-    deadline = time.monotonic() + _REPLY_TIMEOUT
-    while time.monotonic() < deadline:
-        try:
-            return rq.get(timeout=_REPLY_POLL_INTERVAL)
-        except queue.Empty:
-            continue
-    raise TimeoutError("No reply received from Streamlit within the timeout period.")
+    """Wait on *rq* for up to *_REPLY_TIMEOUT* seconds.
+
+    The orchestrator thread is a daemon thread, so it is automatically
+    killed when the main process exits.  A finite timeout is still set so
+    that a closed browser tab eventually releases the thread.
+    """
+    try:
+        return rq.get(timeout=_REPLY_TIMEOUT)
+    except queue.Empty:
+        raise TimeoutError("No reply received from Streamlit within the timeout period.")
 
 
 class StreamlitInterface(HumanInterface):


FaroutYLq (commented):
## Code Review

### Summary

The core refactoring of `_blocking_get()` is logically sound, but there is a **runtime-breaking bug** introduced by the change that will crash the GUI immediately after the orchestrator starts.

---

### Critical Bug: `import time` removed but `time.sleep(0.3)` remains

The PR removes `import time` from the top of `gui.py`, but `time.sleep(0.3)` is still called on line 282 inside `_streamlit_app()`:

```python
# Poll while running with no pending interaction
if ss.get("running") and not ss.get("finished") and ss.get("pending") is None:
    with st.spinner("Analysis in progress…"):
        time.sleep(0.3)   # <-- NameError: name 'time' is not defined
    st.rerun()
```

This code path executes on every Streamlit rerun while the orchestrator is running and there is no pending user interaction — which is the majority of the analysis lifetime. The result is an uncaught `NameError` that crashes the app on the very first polling cycle after the user submits a phenomenon.

The fix is straightforward: restore `import time` at the top of the file. The PR description says it "drops the unused `import time`", but the import is not unused — it is used in `_streamlit_app()`.

---

### The `_blocking_get()` refactoring itself

**Is the simplified version functionally equivalent?** Yes, for the intended use case.

`queue.Queue.get(timeout=N)` in CPython is implemented as a condition variable wait with an absolute deadline. The loop in the old code:

```python
deadline = time.monotonic() + _REPLY_TIMEOUT
while time.monotonic() < deadline:
    try:
        return rq.get(timeout=_REPLY_POLL_INTERVAL)
    except queue.Empty:
        continue
raise TimeoutError(...)
```

is doing exactly what `rq.get(timeout=_REPLY_TIMEOUT)` does internally — it acquires a `threading.Condition`, calls `condition.wait(timeout=remaining)` in a loop to guard against spurious wakeups, and raises `queue.Empty` after the total deadline has passed. The manual re-implementation added no extra capability.

**The GC/daemon-thread argument in the old docstring.** The PR's counter-argument is correct. Python daemon threads are killed unconditionally when the main thread exits — they do not need to reach a checkpoint or poll periodically to be eligible for termination. The old claim that short-interval polling prevented "daemon-thread GC leaks" was incorrect; daemon threads are not garbage-collected in the traditional sense, they are simply killed. The finite `_REPLY_TIMEOUT` is the right mechanism for the "abandoned browser tab" scenario, and it is preserved.

**Signal handling / `KeyboardInterrupt`.** In CPython, signals are only delivered to the main thread. `_blocking_get` runs inside a thread pool executor (via `run_in_executor`) in the orchestrator's daemon thread, so signal delivery behaviour is unchanged between the two implementations. `KeyboardInterrupt` cannot interrupt either version in that thread.

**GIL release.** `queue.Queue.get()` releases the GIL while waiting on the underlying `threading.Condition`. The old `rq.get(timeout=1.0)` in the loop did the same. There is no regression here.

**One subtle difference: `TimeoutError` timing precision.** The old loop could fire the `TimeoutError` up to `_REPLY_POLL_INTERVAL` (1 second) after the true deadline, because each iteration could wait up to 1 second before checking `time.monotonic() < deadline` again. The new single-call version fires within OS timer precision (~10 ms on most platforms). This is strictly better.

**Scenario where polling could have been genuinely needed.** There is no such scenario in this codebase. The one argument for short-interval polling in a blocking wait is to support periodic housekeeping (e.g. checking a `threading.Event` for cooperative cancellation). `_blocking_get` has no such housekeeping, and the daemon-thread termination guarantee makes it unnecessary. If cooperative cancellation were ever added, the correct pattern would be an explicit cancellation token passed into the function, not covert polling.

---

### Minor observations

- The new docstring is cleaner and more accurate. Good improvement.
- Removing `_REPLY_POLL_INTERVAL` eliminates a constant with no callers outside the now-deleted loop. Correct.
- The PR description's test plan is appropriate, though item 2 ("verify closing the browser tab eventually lets the background thread exit") takes up to 1 hour in the worst case. Using a shorter `_REPLY_TIMEOUT` in a dev/test build would make this verifiable in practice.

---

### Required fix before merge

Restore `import time` at the top of `gui.py`. Without this fix the PR introduces a `NameError` at line 282 that breaks the Streamlit GUI for every user during the analysis progress polling loop.

PR #16, opened 2026-03-22
Add post-report follow-up chat (FollowUpChatAgent)

FaroutYLq:
## Summary

- Adds `mtf/agents/followup.py` — `FollowUpChatAgent` that wraps `BaseAgent` with no tools, injects all `SharedMemory` kinds into every query, and accumulates a local conversation history for multi-turn coherence
- Adds `MTFOrchestrator._run_followup_chat()` — opt-in Q&A loop (confirm gate, ask loop, show response) called at the end of `run()` before GPD cleanup
- Updates `docs/architecture/overview.md`: mermaid pipeline diagram extended with a FOLLOW-UP CHAT subgraph; new Phase 4 prose section; file layout updated
- Updates `docs/architecture/agents.md`: new FollowUpChatAgent section with metadata table and multi-turn memory explanation

## Design decisions

- **Single agent, not panel**: reviewer/proposal verdicts are already in SharedMemory and injected into every prompt — no need to re-run the fan-out and debate cycle per question
- **History in task string**: sdk.query() is stateless per call; prior exchanges are prepended as a User/Assistant block so the agent has conversational memory
- **No new MemoryKind**: follow-up exchanges are ephemeral to the session and do not need to survive to future phases or runs

## Test plan

- [ ] Run `pytest tests/test_memory.py tests/test_debate.py tests/test_phases.py` — all pass (MockInterface.confirm returns False by default, skipping the chat loop)
- [ ] Manual smoke test: run `mtf "some phenomenon"`, reach final report, answer y to follow-up prompt, ask a question, verify answer references analysis results, type exit to quit
- [ ] Verify empty-input exit path works (press Enter at question prompt)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -295,7 +295,7 @@ Runs concurrently with `ReviewerAgent` instances inside the review phase. Propos
 |---|---|
 | **Phase** | ④ Follow-up Chat (post-report, optional) |
 | **API** | `sdk.query()` (agentic, no tools) |
-| **Memory context read** | `LITERATURE`, `DEBATE`, `HYPOTHESIS`, `FIT_RESULT`, `REVIEW`, `PROPOSALS`, `USER_FEEDBACK`, `IMAGE_DATA`, `CONVENTIONS`, `PHYSICS_VERDICT`, `FITTING_WARNINGS`, `QUALITATIVE_EVAL` |
+| **Memory context read** | `LITERATURE`, `DEBATE`, `HYPOTHESIS`, `FIT_RESULT`, `REVIEW`, `PROPOSALS`, `USER_FEEDBACK`, `IMAGE_DATA`, `CONVENTIONS`, `PHYSICS_VERDICT`, `FITTING_WARNINGS`, `QUALITATIVE_EVAL`, `DOMAIN_PATTERNS`, `FITTING_SKIPPED`, `TOOLKIT_DIGEST` |
 | **Memory written** | None |
 
 Created by `MTFOrchestrator._run_followup_chat()` after the final report is shown.  A single
--- a/docs/architecture/agents.md
+++ b/docs/architecture/agents.md
@@ -289,6 +289,30 @@ Runs concurrently with `ReviewerAgent` instances inside the review phase. Propos
 
 ---
 
+## FollowUpChatAgent
+
+| | |
+|---|---|
+| **Phase** | ④ Follow-up Chat (post-report, optional) |
+| **API** | `sdk.query()` (agentic, no tools) |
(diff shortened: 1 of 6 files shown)


FaroutYLq:
## Code Review (automated)

### Critical

**`config.model` does not exist** (`mtf/agents/followup.py:47`)
`MTFConfig` has no `.model` field — it uses per-role fields (`reviewer_model`, `debate_model`, etc.). Calling `config.model` raises `AttributeError` at runtime on the first follow-up question.

### High

**Unhandled exceptions from `sdk.query()` propagate out of the loop**
If the API call fails inside `_run_followup_chat()`, the exception escapes `run()` and the already-computed `final_report` is never returned to the caller. A `try/except` around `agent.chat()` with an `interface.show(error)` + `continue` is the right fix.

**History grows without bound**
`_history` accumulates every exchange with no cap. A long session will eventually exceed the model context window and produce an unhandled API error (compounded by the point above).

### Medium

**`_ALL_KINDS` silently omits three `MemoryKind` values**
`DOMAIN_PATTERNS`, `FITTING_SKIPPED`, and `TOOLKIT_DIGEST` are absent. `FITTING_SKIPPED` in particular changes how the reviewer behaves — a follow-up question about missing chi² values would be unanswerable without it.

**No tests**
The `config.model` bug would be caught immediately by a trivial construction test. The existing suite silently skips the new code path because `MockInterface.confirm` defaults to `False`.

### Low

- `_run_followup_chat()` is inside the main `try` block; an error there masks a completed run. It should be outside `try` with its own `try/except`.
- `FollowUpChatAgent.__init__` should accept `gpd_tools: list | None = None` and `gpd: GPDMCPClient | None = None` for consistency with every other agent.
- Mermaid diagram is missing the explicit `CQ -->|"yes"| CA` edge.
- No UI feedback when the user declines the chat (minor UX).

FaroutYLq:
All review issues addressed in the follow-up commit:

- **Critical fixed**: Added `config.followup_model` field to `MTFConfig`; agent now uses it instead of the non-existent `config.model`
- **High fixed**: `agent.chat()` wrapped in `try/except` inside the loop — errors are displayed and the loop continues; `_run_followup_chat()` moved outside the `try/finally` block so chat errors cannot mask a completed run
- **High fixed**: `_history` capped to `_MAX_HISTORY = 10` exchanges (rolling window)
- **Medium fixed**: `_ALL_KINDS` now includes `DOMAIN_PATTERNS`, `FITTING_SKIPPED`, `TOOLKIT_DIGEST`
- **Medium fixed**: Added `tests/test_followup.py` with 10 tests (construction, history accumulation, rolling window, history prepend, skip-on-decline, empty-input exit, quit-keyword exit, API error recovery, answer display)
- **Low fixed**: Constructor now accepts `gpd_tools` and `gpd` params (ignored, for API consistency)
- **Low fixed**: `_run_followup_chat()` moved outside `try/finally`
- **Low fixed**: Mermaid diagram now has explicit `CQ-->|yes|CA` edge and a self-loop on `CA`
- **Nit fixed**: Added `interface.show()` when user declines follow-up chat

PR #20, opened 2026-03-24
feat: citation verification in LiteratureAgent

FaroutYLq:
Adds explicit citation verification step to LiteratureAgent. Agent re-verifies up to 10 citations via search tools, flagging unconfirmed ones [UNVERIFIED]. Adds citation_verification config flag (gates the feature) and citation_verification_max cap (default 10).

**Depends on #19** — merge feat/memory-honesty first.

Motivated by vibe-physics: Claude hallucinates bibliography metadata.

### DIFF
--- a/mtf/agents/literature.py
+++ b/mtf/agents/literature.py
@@ -32,12 +32,13 @@ the experimental phenomenon provided by the user.
 5. When you find a systematic error pattern in a class of papers (wrong conventions,
    missing factors, sign errors), call `add_pattern(category='convention-pitfall', ...)`
    to record it in the cross-session pattern store for future runs.
-6. Citation verification: Before finalising your report, verify every citation in
-   section 4b by calling the search tool again with the exact paper title. Cross-check
-   that the returned author names, publication year, and venue (journal or arXiv ID)
-   match what you plan to report. Flag any citation you cannot confirm as
-   `[UNVERIFIED: <reason>]`. Do not invent or approximate author names — if unsure,
-   omit the author and write only the title and year.
+6. Citation verification: Before finalising your report, verify up to 10 of the most
+   important citations (not all) in section 4b by calling the search tool again with
+   the exact paper title. Cross-check that the returned author names, publication year,
+   and venue (journal or arXiv ID) match what you plan to report. Flag any citation you
+   cannot confirm as `[UNVERIFIED: <reason>]`. Do not invent or approximate author names
+   — if unsure, omit the author and write only the title and year. Skip well-known
+   foundational papers (e.g., BCS theory, Higgs mechanism).
 
 Prefer first-principles hypotheses over phenomenological curve fits. Be precise. All citations must be verified against actual search results — do not rely on memory for author names or publication years."""
 
@@ -49,6 +50,7 @@ class LiteratureAgent(BaseAgent):
         model: str,
         memory: SharedMemory,
         gpd_tools: list[Any] | None = None,
+        config: Any | None = None,
     ) -> None:
         extra_tools: list[Any] = gpd_tools if gpd_tools is not None else []
         super().__init__(
@@ -58,12 +60,24 @@ class LiteratureAgent(BaseAgent):
             memory=memory,
             system_prompt=_SYSTEM_PROMPT,
         )
+        self._config = config
 
     async def investigate(self, phenomenon: str) -> str:
+        verification_note = ""
+        if self._config is None or getattr(self._config, "citation_verification", True):
+            verification_note = (
+                "\n\nCITATION VERIFICATION (max 10 citations): Before finalising section 4b, "
+                "verify up to 10 of the most important citations by calling the search tool "
+                "again with the exact paper title. Cross-check that returned author names, "
+                "year, and venue match what you plan to report. Flag unverified ones as "
+                "[UNVERIFIED: <reason>]. Skip re-verification for well-known foundational "
+                "papers (e.g., BCS theory, Higgs mechanism)."
+            )
         task = (
             f"Investigate the following experimental phenomenon and search for "
             f"relevant literature:\n\n{phenomenon}\n\n"
             "Produce a comprehensive literature report with hypotheses."
+            + verification_note
         )
         report = await self._query(
             task,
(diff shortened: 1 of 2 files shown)


FaroutYLq (commented):
## Code Review: feat/literature-citations

### Summary

The PR bundles three distinct changes: (1) a citation-verification step added to `LiteratureAgent`'s system prompt, (2) a honesty/convention-enforcement injection added to `BaseAgent._build_prompt()`, and (3) a navigable index prepended to `SharedMemory.format_context()` when more than 3 entries are present. These came from separate commits on `feat/memory-honesty`, and the PR title says only "citation verification." The scope mismatch needs to be addressed before merge.

---

### 1. System prompt change — citation verification (step 6)

**Integration with the 5-step structure:** The new step 6 is syntactically clean and reads naturally after step 5 (pattern recording). However, the placement is slightly off logically: step 4b is where citations are assembled, so the instruction to verify them would be clearer if it appeared as a sub-step of 4b rather than a new top-level step that happens after pattern recording. Currently a reader has to jump back to understand what "section 4b" refers to.

**`[UNVERIFIED]` tag definition:** The tag format `[UNVERIFIED: <reason>]` is defined inline in step 6 and is clear enough. No issues here.

**Instruction specificity — what to call and what to compare:** The instruction says "call the search tool again with the exact paper title." This is underspecified in two ways:

- The agent has two search tools (`arxiv_search` and `semantic_search`). The instruction does not say which to use for verification, or whether both should be tried when the first returns no match. A model following the prompt literally may pick one arbitrarily.
- The instruction says to cross-check "author names, publication year, and venue (journal or arXiv ID)" but does not say what to do when there is a partial match (e.g., title found, year matches, but first author differs). "Flag as UNVERIFIED" applies only when the agent "cannot confirm" — a partial mismatch is ambiguous.

**Recommendation (non-blocking):** Tighten step 6 to name both tools explicitly and define the partial-match rule (e.g., flag if any of the three fields differs by more than a plausible formatting variation).

---

### 2. Tool availability

The default `literature_model` in `config.py` is `claude-haiku-4-5-20251001`. The citation verification step requires the agent to make additional tool calls within the same turn. Both `make_arxiv_search_tool()` and `make_semantic_search_tool()` are always injected via `LiteratureAgent.__init__`, so the tools themselves are always available. However, the instruction is in the system prompt, not a separate post-processing step, so the agent must self-initiate re-verification before finalising the report. There is no fallback if the agent omits step 6 under a heavy context load. This is a prompt-compliance risk rather than a hard failure, but worth noting: haiku models under long contexts can silently skip steps near the end of the system prompt.

---

### 3. Performance impact — citation storm

**BLOCKER:** There is no guidance in the prompt on how many citations require verification. A thorough literature agent might surface 20-40 papers in section 4b. Step 6 says "verify every citation," which means 20-40 additional search calls per agent, multiplied by `n_literature` (default 2). With `n_literature=2`, a single literature phase could generate 80+ verification calls. There is no semaphore guarding these calls (the semaphore in `config.py` only applies to `fitting_semaphore_limit`). This risks rate-limit errors and significantly increases latency and cost.

Suggest either: (a) limiting verification to citations where the agent has low confidence (e.g., papers retrieved only from model memory, not from the current tool results), or (b) adding a `citation_verification_max: int` config field, or (c) capping step 6 explicitly ("verify the top N citations by citation count or centrality").

---

### 4. `citation_verification` config field — not wired to code

**BLOCKER:** `config.citation_verification` is declared in `MTFConfig` but is never read anywhere in the codebase (confirmed by grepping all of `mtf/`). When `citation_verification=False`, verification still happens unconditionally because it is baked into the static `_SYSTEM_PROMPT` string. The field is dead code.

The same applies to `honesty_enforcement` and `memory_index` introduced in the same diff: both are declared in `config.py` but `base.py` always appends `_HONESTY_REMINDER` unconditionally and `memory.py` always prepends the index unconditionally, with no reference to the config fields. This is inconsistent with every other boolean flag in `MTFConfig`, all of which actually gate the behaviour they describe.

The fix is either to wire the flags (pass config into `BaseAgent` and `SharedMemory`, check the flags before injecting), or to document clearly in the dataclass that these fields are reserved/aspirational. The CLAUDE.md invariant added in this PR ("Never remove these") makes the problem worse — it permanently bakes in unconditional behaviour while leaving the user-facing off-switch inoperative.

---

### 5. Placement of step 6 relative to step 4

As noted above, step 6 comes after the pattern-recording step (5). From a workflow perspective, the natural order is: search → structure report with citations → **verify citations while report is being assembled** → record error patterns. Having verification come after pattern recording means the agent might record patterns for papers it later discovers are unverified. Suggested order: 1 → 2 → 3 → 4 (with verification woven into 4b) → 5.

---

### 6. Branch base — PR scope

**BLOCKER:** This PR is based on `feat/memory-honesty`, not `main`. The diff against `main` includes:

- `mtf/agents/base.py`: `_HONESTY_REMINDER`, `_CONVENTION_REMINDER`, modified `_build_prompt()`
- `mtf/memory.py`: `format_index()`, index injection in `format_context()`, `MemoryKind.PHENOMENON`
- `mtf/orchestrator.py`: `PHENOMENON` anchoring at run start
- `mtf/config.py`: `honesty_enforcement`, `memory_index` fields
- `tests/test_memory.py`: 5 new test functions for the memory changes
- `CLAUDE.md`: documentation for all of the above

None of these are "citation verification." They should be in a separate PR (presumably `feat/memory-honesty`) that gets merged first, or this PR should be rebased onto `main` and scoped to only the citation-verification change. Reviewing mixed scopes makes it hard to assess individual changes and complicates bisecting if a regression is introduced.

---

### Minor issues

- The trailing sentence added to `_SYSTEM_PROMPT` ("All citations must be verified against actual search results — do not rely on memory for author names or publication years.") duplicates the instruction already stated in step 6. One or the other should be removed.
- `format_index()` and the inline index in `format_context()` use different preview lengths (100 vs 80 characters) for no stated reason. This is confusing for a reader comparing the two methods.
- The test `test_format_context_prepends_index_when_more_than_3_entries` asserts `ctx.count("[LITERATURE]") == 8` (4 in index + 4 full). This is a fragile assertion: if the index format changes to omit the kind tag, the count breaks. Better to assert that the full-entries section contains all 4 items independently of the index section.

---

### Verdict: REQUEST_CHANGES

Three blockers must be addressed before merge:

1. **`citation_verification` (and `honesty_enforcement` / `memory_index`) are not wired** — the config flags do not gate the behaviour they describe.
2. **No citation-count cap** — "verify every citation" can cause a rate-limit cascade for agents that surface large bibliographies.
3. **PR scope is incorrect** — the diff includes `feat/memory-honesty` changes that should be merged separately; this PR should be rebased to contain only the citation-verification work.

PR #21, opened 2026-03-24
feat: fitting result integrity checks against fabrication

FaroutYLq:
## Summary
- Wraps lmfit minimize()/Model.fit() in sentinels to detect whether any real optimizer call was made
- Validates result dict post-exec: chi² non-negative, parameters non-empty
- Adds MemoryKind.INTEGRITY_WARNING (separate from PHYSICS_VERDICT) for fabrication flags
- Integrity warnings surfaced to ReviewerAgent via extra_kinds
- fitting_result_integrity_check config flag gates the feature
- Adds functools.wraps on sentinel; adds 5 unit tests

## Motivation
From Schwartz's vibe-physics paper: Claude adjusted parameters to make plots match rather than running real computations.

## Merge order
Merge after #19 and #20. Conflicts with #19 on memory.py (both add MemoryKind values) — resolve by keeping both PHENOMENON and INTEGRITY_WARNING.

🤖 Generated with Claude Code

### DIFF
--- a/mtf/agents/fitting.py
+++ b/mtf/agents/fitting.py
@@ -29,16 +29,23 @@ experimental physics. Given a hypothesis and experimental data, you:
 4. Identify what data and model functions are needed from the toolkit.
 5. Write Python code using lmfit/numpy/scipy to fit the data, following the protocol
    retrieved above and incorporating the correct conventions.
-6. In the result dict, always include:
+6. CRITICAL — anti-fabrication: Do NOT hardcode or manually adjust result values.
+   Your result dict MUST be populated directly from lmfit's MinimizerResult:
+   copy `result.chisqr`, `result.redchi`, and each `params[name].value` /
+   `params[name].stderr` directly from the object returned by `minimize()` or
+   `Model.fit()`. Never write `result = {"chi_squared": <manually chosen number>}`.
+   If the optimizer fails to converge, report that failure honestly — do not
+   substitute adjusted numbers to make the fit appear successful.
+7. In the result dict, always include:
    - 'protocol_followed': name of the protocol retrieved via get_protocol
    - 'physical_parameter_ranges': dict mapping each parameter name to whether its
      best-fit value falls within expected physical bounds (True/False + note)
    - 'protocol_checkpoints_satisfied': list of protocol checkpoint names that passed
-7. Report fit quality (chi-squared, reduced chi-squared, residuals), best-fit
+8. Report fit quality (chi-squared, reduced chi-squared, residuals), best-fit
    parameters with uncertainties, and an assessment of whether the hypothesis is
    supported. Chi-squared is a necessary metric but NOT the sole quality criterion;
    physical correctness (parameter bounds, limiting cases, symmetry) matters more.
-8. If your fitting code fails to converge or produces unphysical parameters, call
+9. If your fitting code fails to converge or produces unphysical parameters, call
    `add_pattern(category='convergence-issue', ...)` to record the pattern for future
    runs on this domain.
 
@@ -166,6 +173,17 @@ class FittingAgent(BaseAgent):
                     break  # No violation — proceed to exec
 
         fit_output = run_fitting_code(code, self._toolkit.all_data())
+
+        # Surface integrity warnings in memory so reviewers see them
+        if self._config is not None and fit_output.get("integrity_warnings"):
+            for w in fit_output["integrity_warnings"]:
+                self._memory.add(
+                    MemoryKind.PHYSICS_VERDICT,
+                    w,
+                    source="fitting_integrity_check",
+                    hypothesis=hypothesis[:200],
+                )
+
         report = f"Hypothesis: {hypothesis}\n\nFit output: {fit_output}"
         self._memory.add(
             MemoryKind.FIT_RESULT,
--- a/mtf/agents/fitting.py
+++ b/mtf/agents/fitting.py
@@ -172,13 +172,14 @@ class FittingAgent(BaseAgent):
                 else:
                     break  # No violation — proceed to exec
 
-        fit_output = run_fitting_code(code, self._toolkit.all_data())
+        enabled = self._config.fitting_result_integrity_check if self._config is not None else True
+        fit_output = run_fitting_code(code, self._toolkit.all_data(), integrity_check=enabled)
 
         # Surface integrity warnings in memory so reviewers see them
         if self._config is not None and fit_output.get("integrity_warnings"):
             for w in fit_output["integrity_warnings"]:
                 self._memory.add(
-                    MemoryKind.PHYSICS_VERDICT,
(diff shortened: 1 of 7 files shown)


FaroutYLq (commented):
## Code Review: feat/fitting-integrity

Good motivation — the vibe-physics failure mode (hardcoding results to match plots) is real and worth defending against. The PR is small, focused, and introduces a useful sentinel pattern. Several issues need to be addressed before merging.

---

### 1. `_make_sentinel_minimize` — signature preservation

**Suggestion.** The wrapper uses `*args, **kwargs` which correctly forwards all positional and keyword arguments, including `minimize(fcn, params, method='leastsq')`. That part is fine. However, the wrapper does not preserve `__name__`, `__doc__`, or `__module__`. While not a runtime error, this can confuse `inspect`-based tools or error messages. Recommend adding `@functools.wraps(_real_minimize)` inside the factory.

---

### 2. `_make_sentinel_model` — usage as a class replacement

The function returns the `SentinelModel` **class** and is injected as `"Model": _make_sentinel_model(optimizer_called)`. This is correct — exec'd code does `Model(func)` which calls `SentinelModel.__init__` (inherited from `lmfit.Model`), then `.fit()` fires the sentinel. The subclass does not alter `Model.__init__`'s signature, so construction is unaffected.

One concern: if the agent writes `class MyModel(Model):` inside the exec'd code, `Model` in that namespace resolves to `SentinelModel`, so `MyModel` inherits the sentinel — this actually works correctly. No issue here.

Minor: the PR description says `_make_sentinel_model_class` but the actual function is `_make_sentinel_model`. Rename one to match.

---

### 3. `optimizer_called` list flag — multiple minimize() calls

**Non-blocker.** The `[False]` closure pattern is idiomatic Python for mutable closure state. Setting `called_flag[0] = True` (not toggling) is correct for the "was it ever called?" semantic. This handles multiple `minimize()` calls correctly.

---

### 4. Residuals cross-check tolerance — false positives on weighted fits

**Blocker.** The 5% tolerance check uses:

```python
abs(recomputed - declared) / max(abs(declared), 1e-10) > 0.05
```

`lmfit`'s `chisqr` is defined as `sum((residuals / sigma)^2)` when uncertainties are provided. The recomputed value is `sum(residuals^2)`. If the agent stores raw (unweighted) residuals but reports the correctly weighted chi-squared, the cross-check fires a false-positive warning on every well-executed weighted fit.

The `max(abs(declared), 1e-10)` guard does prevent division by zero when `chi_squared == 0`. But the sigma-weighting mismatch is a more common source of false positives that will erode trust in the warning system. The check should either (a) be documented as applicable only to unweighted fits, (b) use a much more conservative threshold (e.g., factor-of-10 discrepancy), or (c) be removed until a robust weighted vs. unweighted detection strategy is in place.

---

### 5. Integrity warnings stored as `PHYSICS_VERDICT` — semantic mismatch

**Blocker.** `MemoryKind.PHYSICS_VERDICT` is documented in `memory.py` as "structured check results from gpd-verification" and is populated by GPD's `run_check`, `dimensional_check`, and `limiting_case_check` calls. The PR repurposes this kind for runtime execution integrity warnings, which have a completely different provenance.

`ReviewerAgent` reads `PHYSICS_VERDICT` in its `extra_kinds` and is designed to interpret GPD domain-physics check results. Injecting fabrication warnings into that stream risks the reviewer misinterpreting "no optimizer was called" as a domain physics verdict (or a GPD check result being dismissed as a fabrication warning).

A new `MemoryKind.INTEGRITY_WARNING = "integrity_warning"` would be cleaner and is a one-line addition to `memory.py`. This keeps the semantic streams separate and allows reviewers (and future display logic) to treat them differently.

---

### 6. `fitting_result_integrity_check` config field — never consulted

**Blocker.** `MTFConfig.fitting_result_integrity_check = True` is added to `config.py` but `run_fitting_code()` in `fitting_tools.py` always runs `_validate_result()` unconditionally — it does not receive or check this flag. The `fit()` method in `fitting.py` checks `self._config is not None` before writing warnings, but never checks `self._config.fitting_result_integrity_check`.

This is dead config. Either (a) wire the flag through so `run_fitting_code()` accepts an optional `integrity_check: bool = True` parameter and the `fit()` method passes `self._config.fitting_result_integrity_check`, or (b) remove the field from config if the check is intentionally always-on. The current state is a lie: setting `fitting_result_integrity_check = False` has no effect.

---

### 7. Anti-fabrication prompt — scipy.optimize bypass not covered

**Suggestion.** The new item 6 in `_SYSTEM_PROMPT` correctly calls out lmfit's `minimize()` and `Model.fit()`. However, the exec namespace also exposes `optimize` (which is `scipy.optimize`). An agent using `scipy.optimize.curve_fit` or `scipy.optimize.minimize` directly would bypass both sentinels entirely and produce no integrity warning, even while performing a legitimate fit.

The prompt says "Do NOT hardcode or manually adjust result values" but does not mention that scipy is also acceptable, nor that using it directly bypasses the integrity check. Consider either: (a) adding sentinels on `optimize.minimize` and `optimize.curve_fit` as well, or (b) updating the prompt to require lmfit exclusively. The gap means an agent can legitimately circumvent the sentinel without fabricating anything.

---

### 8. Tests — no automated coverage for the new functionality

**Blocker.** The PR's test plan lists manual verification steps but adds zero automated tests. The existing test suite (`test_memory.py`, `test_debate.py`, `test_phases.py`) has no coverage for:

- `_validate_result` directly: negative chi-squared, empty parameters dict, residual mismatch, clean result
- `run_fitting_code` with fabricated result (no optimizer call) → `integrity_warnings` present in output
- `run_fitting_code` with legitimate lmfit `minimize()` call → no `integrity_warnings`
- `FittingAgent.fit()` writing `PHYSICS_VERDICT` (or the proposed `INTEGRITY_WARNING`) entries when warnings exist

These are pure unit tests requiring no API key and would fit naturally alongside the existing phase tests. Without them, a future refactor of `run_fitting_code` or `_validate_result` could silently break the integrity checks with no CI signal.

---

### Summary

| # | Issue | Severity |
|---|---|---|
| 4 | False-positive warnings on sigma-weighted fits | Blocker |
| 5 | Fabrication warnings pollute `PHYSICS_VERDICT` semantic stream | Blocker |
| 6 | `fitting_result_integrity_check` config field never consulted (dead config) | Blocker |
| 8 | No automated tests for sentinel / validation logic | Blocker |
| 1 | `functools.wraps` missing on sentinel wrapper | Suggestion |
| 7 | scipy.optimize bypass not covered by sentinel or prompt | Suggestion |
| 2/3 | Sentinel class usage and `[False]` flag pattern are correct | OK |

**Verdict: REQUEST_CHANGES**

The core idea is sound and the sentinel mechanism itself is correctly implemented. Address the four blockers — particularly the weighted-residuals false-positive risk, the `MemoryKind` mismatch, the dead config field, and the missing tests — before merging.

PR #22, opened 2026-03-24
feat: reviewer second-pass loop and multi-model diversity

FaroutYLq:
## Summary
- Adds exhaustive review requirement to ReviewerAgent system prompt
- Adds structural second-pass verification loop: reviewer_verification_passes > 1 triggers re-check
- Adds multi-model cycling via reviewer_models config field
- Config fields placed next to n_reviewer; direct attribute access (no getattr)
- Adds 4 unit tests

## Motivation
From Schwartz's vibe-physics paper: Claude stops after finding first error; different models catch different errors.

## Merge order
Independent of #21 and #23. Merge after #19/#20. Conflicts on reviewer.py if merged after #21 (both touch it) — resolve by keeping both changes.

🤖 Generated with Claude Code

### DIFF
--- a/mtf/agents/reviewer.py
+++ b/mtf/agents/reviewer.py
@@ -48,7 +48,14 @@ If `[FITTING_SKIPPED]` appears in your context, the fitting phase was skipped an
 - Do not rank by chi² — rank by: (1) physics check results on theoretical arguments,
   (2) parsimony, (3) first-principles basis.
 - Your recommendation should focus on what data collection would be needed to upgrade
-  qualitative assessments to quantitative fits."""
+  qualitative assessments to quantitative fits.
+
+EXHAUSTIVE REVIEW REQUIREMENT: After completing steps 1-9 above, re-read your \
+entire review from the start. Do NOT stop after identifying the most prominent problem. \
+Enumerate ALL issues found, including minor concerns, under an \
+'Additional concerns:' section at the end of your report. A review that identifies \
+three issues is more valuable than one that stops at the first. Only conclude when \
+you have nothing new to add."""
 
 
 class ReviewerAgent(BaseAgent):
--- a/mtf/config.py
+++ b/mtf/config.py
@@ -14,6 +14,8 @@ class MTFConfig:
     n_fitting: int = 2
     n_qualitative: int = 2
     n_reviewer: int = 2
+    reviewer_verification_passes: int = 1  # >1 enables structural "check again" loop
+    reviewer_models: list[str] = field(default_factory=list)  # cycles models for diversity
 
     # Model selection
     literature_model: str = "claude-haiku-4-5-20251001"
@@ -53,9 +55,6 @@ class MTFConfig:
     fitting_result_integrity_check: bool = True  # detect hardcoded/fabricated fit results
 
     # Proposal agent
-    reviewer_verification_passes: int = 1  # >1 enables structural "check again" loop
-    reviewer_models: list[str] = field(default_factory=list)  # cycles models for diversity
-
     n_proposal: int = 2
     proposal_model: str = "claude-sonnet-4-6"
     followup_model: str = "claude-sonnet-4-6"
--- a/mtf/config.py
+++ b/mtf/config.py
@@ -53,6 +53,9 @@ class MTFConfig:
     fitting_result_integrity_check: bool = True  # detect hardcoded/fabricated fit results
 
     # Proposal agent
+    reviewer_verification_passes: int = 1  # >1 enables structural "check again" loop
+    reviewer_models: list[str] = field(default_factory=list)  # cycles models for diversity
+
     n_proposal: int = 2
     proposal_model: str = "claude-sonnet-4-6"
     followup_model: str = "claude-sonnet-4-6"
--- a/mtf/phases/review_phase.py
+++ b/mtf/phases/review_phase.py
@@ -84,10 +84,16 @@ async def run_review_phase(
         title="MTF: Review",
     )
 
+    def _reviewer_model(i: int) -> str:
+        models = getattr(config, "reviewer_models", [])
+        if models:
+            return models[i % len(models)]
+        return config.reviewer_model
+
     agents = [
         ReviewerAgent(
             agent_id=f"rev-{i}",
-            model=config.reviewer_model,
+            model=_reviewer_model(i),
             memory=memory,
             gpd_tools=gpd_tools,
         )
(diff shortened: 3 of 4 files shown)


FaroutYLq (commented):
## Code Review: feat/reviewer-exhaustive-v2

### Summary

The PR adds three things: (1) an exhaustive-review instruction paragraph to `_SYSTEM_PROMPT`, (2) a second-pass verification loop in `run_review_phase()`, and (3) multi-model cycling via `reviewer_models`. The intent is sound. There are a few correctness issues and one blocker.

---

### 1. Second-pass loop — `asyncio.gather()` correctness

**Status: Correct, but needs a comment.**

`asyncio.gather(*(agent._query(task, extra_kinds=(...)) for agent, task in zip(agents, second_pass_tasks)))` is properly awaited. The zip pairing is correct: `agents[i]` is matched with `second_pass_tasks[i]`, which contains `reviewer_reports[i]` from the previous pass. The `extra_kinds` tuple in the second pass is identical to the one in `ReviewerAgent.review()` — that is the right choice so agents have full context.

`pass_num` is used as `pass_num + 2` in the f-string heading (pass 1 = initial, pass 2 = first retry), which is arithmetically correct.

**Suggestion:** Add a one-line comment explaining that `agents[i]` corresponds to `reviewer_reports[i]` and that `reviewer_reports` is reassigned each iteration to chain passes.

---

### 2. `getattr(config, "reviewer_verification_passes", 1)` — BLOCKER

**Status: Bug / unnecessary defensive code that masks a real problem.**

`reviewer_verification_passes` is now a declared field of `MTFConfig` with a default of `1`. Using `getattr(..., 1)` instead of `config.reviewer_verification_passes` is unnecessary and silently swallows any `AttributeError` that would otherwise signal a misconfiguration (e.g., a caller passing a different config class, a serialisation round-trip that drops the field). mypy will not catch that the attribute is being accessed through `getattr`; the static type safety the dataclass provides is bypassed.

The same issue applies to `getattr(config, "reviewer_models", [])` inside `_reviewer_model()`.

**Fix:** Use `config.reviewer_verification_passes` and `config.reviewer_models` directly. The dataclass guarantees the fields exist at construction time. If forward-compatibility with older pickled configs is genuinely needed, document that explicitly and centralise the migration logic rather than sprinkling `getattr` guards.

---

### 3. `_reviewer_model` inner function — correctness and performance

**Status: Functionally correct; one edge case.**

Defining `_reviewer_model` as an inner function of `run_review_phase()` is fine — it is created once per call and has negligible overhead. No closure state is captured other than `config`.

**Edge case:** `getattr(config, "reviewer_models", [])` (addressed above) can return `None` if a caller explicitly sets `reviewer_models=None`. The guard `if models:` treats `None` as falsy and falls through to `config.reviewer_model` — correct in practice, but fragile. With the field typed as `list[str]`, mypy will flag `None` as a type error, so direct attribute access plus the existing type annotation is the right safeguard.

---

### 4. Second-pass agent reuse — documentation gap

**Status: Correct, but undocumented.**

`BaseAgent._query()` is stateless: no internal history, no mutable fields modified per call. Reusing the same `agents` list across passes is safe. However, there is no comment noting this invariant, which means a future contributor who adds per-call state to `BaseAgent` could unknowingly break the second-pass assumption.

**Suggestion:** Add a one-line comment at the second-pass block: `# Agents are stateless between calls; safe to reuse.`

---

### 5. Exhaustive requirement in `_SYSTEM_PROMPT` — heading mismatch

**Status: Mostly correct; one phrasing concern.**

The new paragraph is appended with a backslash line continuation inside a triple-quoted string, which is valid Python. The resulting text reads correctly.

**Concern:** The system prompt instructs the model to use an `Additional concerns:` section heading. The second-pass task prompt in `review_phase.py` uses `Additional concerns (pass {pass_num + 2}):` as the expected heading. These two strings are inconsistent: the system prompt instructs one heading, the task prompt expects a versioned heading. When `DebateEngine.synthesize()` later processes chained reports, mixed headings make structured parsing unreliable if any downstream code looks for these headings.

**Fix:** Either drop the specific heading name from the system prompt (let the task prompt control it), or make the system prompt match the versioned heading pattern.

---

### 6. `reviewer_models: list[str] = field(default_factory=list)`

**Status: Correct.**

`field(default_factory=list)` is the canonical way to declare a mutable list default in a dataclass. It avoids the shared-mutable-default pitfall. No issue here.

---

### 7. `MemoryKind` imports in `review_phase.py`

**Status: Correct.**

`MemoryKind` is already imported via `from mtf.memory import MemoryKind, SharedMemory` at line 13 of `review_phase.py`. All nine `MemoryKind` values used in the second-pass `extra_kinds` tuple (`LITERATURE`, `DEBATE`, `FIT_RESULT`, `USER_FEEDBACK`, `IMAGE_DATA`, `CONVENTIONS`, `PHYSICS_VERDICT`, `QUALITATIVE_EVAL`, `FITTING_SKIPPED`) are defined in `memory.py`. No missing imports.

---

### 8. Test coverage — BLOCKER

**Status: Insufficient.**

The PR test plan lists manual steps but ships zero new automated tests. The existing `test_phases.py` covers only `run_literature_phase`. There are no tests for `run_review_phase` at all, let alone for the new second-pass path. Specifically:

- The default path (`reviewer_verification_passes=1`) is not exercised by any test. A regression in the base reviewer fan-out would go undetected.
- The second-pass path (`reviewer_verification_passes=2`) is not tested. Because `reviewer_reports` is reassigned inside the loop, a subtle bug (e.g., passing `agents` and `second_pass_tasks` in mismatched order) would not be caught.
- `_reviewer_model` cycling (`reviewer_models=["a","b"]`, `n_reviewer=3` producing `["a","b","a"]`) is not tested.

**Minimum required additions:**
1. A test patching `ReviewerAgent._query` to confirm the second-pass task string contains the first-pass report text when `reviewer_verification_passes=2`.
2. A test verifying `_reviewer_model` cycling produces the expected model sequence.
3. A test confirming the default path (`reviewer_verification_passes=1`) does not enter the loop.

---

### 9. Minor: `config.py` field placement

The two new fields (`reviewer_verification_passes`, `reviewer_models`) are inserted under the `# Proposal agent` comment block, which is semantically incorrect — they govern the reviewer, not proposals. They should be placed near `n_reviewer` and `reviewer_model`. Not a functional issue, but confusing to readers of the config.

---

## Verdict: REQUEST_CHANGES

**Blockers (must fix before merge):**
1. Replace `getattr(config, "reviewer_verification_passes", 1)` and `getattr(config, "reviewer_models", [])` with direct attribute access (item 2).
2. Add at minimum the three automated tests described in item 8.

**Suggestions (non-blocking):**
- Reconcile the `Additional concerns:` vs `Additional concerns (pass N):` heading mismatch (item 5).
- Move the two new config fields to the reviewer section of `config.py` (item 9).
- Add the two one-line comments noted in items 1 and 4.

PR #23, opened 2026-03-24
feat: debate anti-consensus and follow-up pressure resistance

FaroutYLq:
## Summary
- DebateEngine anti-consensus paragraph scoped to non-proposals phases only (proposals phase still deduplicates/merges)
- FollowUpChatAgent pressure resistance anchored to existing analysis evidence
- Fixes pre-existing test_gpd_integration.py assertion (== 4 → == 3)

## Motivation
From Schwartz's vibe-physics paper: Claude gives the answer users seem to want under pressure.

## Merge order
Independent. Merge after #19/#20. No expected conflicts.

🤖 Generated with Claude Code

### DIFF
--- a/mtf/agents/followup.py
+++ b/mtf/agents/followup.py
@@ -17,6 +17,13 @@ candidate hypotheses, fitting results, reviewer verdicts, and proposed measureme
 Answer follow-up questions clearly and precisely. Reference specific results, numbers, or \
 verdicts from the analysis where relevant. If the user asks about something outside the \
 scope of the analysis, say so honestly rather than speculating beyond what was established.
+
+PRESSURE RESISTANCE: If the user pushes back on a conclusion from the analysis, \
+re-evaluate the evidence independently rather than accommodating their preference. \
+Changing a position requires new evidence or a logical argument — user insistence alone \
+is not sufficient reason to revise a verdict. If the original conclusion was well-supported, \
+defend it with specific evidence from the analysis. It is scientifically appropriate to \
+disagree with the user.
 """
 
 # All memory kinds that are meaningful for follow-up context.
--- a/mtf/agents/followup.py
+++ b/mtf/agents/followup.py
@@ -19,11 +19,12 @@ verdicts from the analysis where relevant. If the user asks about something outs
 scope of the analysis, say so honestly rather than speculating beyond what was established.
 
 PRESSURE RESISTANCE: If the user pushes back on a conclusion from the analysis, \
-re-evaluate the evidence independently rather than accommodating their preference. \
-Changing a position requires new evidence or a logical argument — user insistence alone \
-is not sufficient reason to revise a verdict. If the original conclusion was well-supported, \
-defend it with specific evidence from the analysis. It is scientifically appropriate to \
-disagree with the user.
+review the specific evidence in the analysis that supports the original conclusion \
+rather than accommodating their preference. Changing a position requires new evidence \
+or a logical argument — user insistence alone is not sufficient reason to revise a \
+verdict. If the original conclusion was well-supported by the analysis, cite the \
+specific results, fit parameters, or reviewer verdicts that justify it. It is \
+scientifically appropriate to disagree with the user.
 """
 
 # All memory kinds that are meaningful for follow-up context.
--- a/mtf/debate.py
+++ b/mtf/debate.py
@@ -57,8 +57,13 @@ class DebateEngine:
         base_system = (
             f"You are a synthesis engine for the {phase} phase of a multi-agent "
             "physics research assistant. Your job is to produce a single coherent "
-            "synthesis of the provided agent reports, resolving contradictions and "
-            "highlighting the strongest hypotheses or conclusions. Be concise and precise."
+            "synthesis of the provided agent reports, highlighting the strongest "
+            "hypotheses or conclusions. Be concise and precise.\n\n"
+            "IMPORTANT — preserve genuine disagreement: If agents reach contradictory "
+            "conclusions, present both views clearly and explain the evidential tension "
+            "rather than smoothing them into false consensus. Use phrases like "
+            "'agents agree' or 'consensus' only when reports genuinely agree. "
+            "Unresolved contradictions are scientifically valuable — do not suppress them."
         )
 
         if phase in ("fitting", "review"):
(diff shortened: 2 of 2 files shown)


FaroutYLq (commented):
## Code Review: feat: debate anti-consensus and follow-up pressure resistance

### Summary

Two targeted changes: (1) `DebateEngine.synthesize()` replaces "resolving contradictions" with a preserve-disagreement instruction, and (2) `FollowUpChatAgent._SYSTEM_PROMPT` gains a pressure-resistance paragraph. Both address the "vibe-physics" failure mode where the model gives users the answer they seem to want. The motivation is sound. Several details need attention before merging.

---

### 1. "Resolving contradictions" removal — correctness across all phases

**`debate.py` lines 57–62 (old) → 57–69 (new)**

The removal is mostly correct, but the blanket replacement has a phase-specific problem.

For `literature` and `review` phases: genuine contradictions (agent A says hypothesis X is supported, agent B says it is speculative) represent real epistemic disagreement and should be preserved. The new phrasing is an improvement here.

For `proposals` phase: the `proposals_criterion` block (lines 80–88) already instructs the synthesizer to "remove redundant proposals, merge overlapping ones." Two agents proposing the same measurement with different priority labels are not epistemically contradicting each other — they are duplicating. The anti-consensus instruction ("present both views clearly and explain the evidential tension") now sits in `base_system`, which is prepended before every phase-specific criterion including `proposals_criterion`. A synthesizer that sees both instructions simultaneously may hedge between merging and preserving, producing a bloated list instead of the intended deduplicated output.

**Blocker**: The anti-consensus paragraph should either be excluded from the proposals phase or explicitly scoped. The simplest fix is to condition the new paragraph on the phase inside the `base_system` construction, or rewrite the paragraph to say "preserve disagreement about *conclusions and evidence*, not about proposal priority or duplication."

---

### 2. Anti-consensus instruction framing

**`debate.py` lines 63–68 (new)**

> "Use phrases like 'agents agree' or 'consensus' only when reports genuinely agree."

This is a meta-instruction about which phrases to use, not about what reasoning to perform. The model may comply literally — avoiding the phrase "agents agree" even when agents do agree — without actually preserving the substance of disagreement in the logical structure of its output. A more effective instruction would focus on the output structure: e.g. "When agents reach different conclusions, include a 'Points of disagreement' subsection that states each position and the evidence behind it." Phrase-policing is a weak proxy for the intended behavior.

**Suggestion** (non-blocking): Reframe the instruction around structure rather than vocabulary. Phrase-avoidance instructions are easy to satisfy superficially.

---

### 3. FollowUpChatAgent — "re-evaluate the evidence independently"

**`agents/followup.py` lines 22–27 (new)**

> "re-evaluate the evidence independently rather than accommodating their preference"

The follow-up agent has no tools and answers entirely from `SharedMemory` context injected via `BaseAgent._build_prompt()`. It cannot fetch new data, re-run fits, or query literature. "Re-evaluate independently" could be read by the model as a license to reason beyond the stored results — generating new analysis that was never validated by the pipeline. This risks introducing hallucinated conclusions under the banner of "independent evaluation."

The existing instruction on line 18 ("say so honestly rather than speculating beyond what was established") already partially guards against this, but the new paragraph could undermine it.

**Blocker**: Replace "re-evaluate the evidence independently" with something like "review the specific evidence in the analysis that supports the original conclusion." This anchors the response to recorded results rather than implying the agent should generate new conclusions.

---

### 4. Tone consistency with the existing "outside scope" instruction

**`agents/followup.py` lines 17–27**

The existing instruction (lines 17–19) is measured and acknowledges limits. The new paragraph is assertive in tone ("It is scientifically appropriate to disagree with the user"). These are compatible in intent but the abruptness of the new paragraph reads as an afterthought. The phrase "defend it with specific evidence from the analysis" is appropriate — it anchors pushback to actual recorded results. The overall integration is acceptable once the "re-evaluate independently" language is fixed.

**Suggestion** (non-blocking): Consider a transitional sentence such as "When the user challenges a specific result from the analysis, ..." rather than the generic "If the user pushes back on a conclusion from the analysis."

---

### 5. No config flag — is that appropriate?

The PR adds no `debate_preserve_disagreement: bool` or `followup_pressure_resistance: bool` flag. Compare `honesty_enforcement: bool = True` and `memory_index: bool = True` in `config.py`, which show the project already uses flags to make always-on honesty behaviors configurable.

The case for a flag here is not primarily about disabling the behavior — it is about testability: tests can set the flag to `False` and assert on an exact `system` string without the new paragraph, isolating what changed. Without a flag, any future edit to the paragraph text requires updating every assertion that captures the system prompt.

**Suggestion** (non-blocking but consistent with project conventions): Add `debate_preserve_disagreement: bool = True` and `followup_pressure_resistance: bool = True` to `MTFConfig`, mirroring `honesty_enforcement`. Zero runtime cost; enables targeted test isolation.

---

### 6. Test coverage

**`tests/test_debate.py`**: No test verifies that the new paragraph appears in the `system` argument passed to `messages.create()`. No test verifies the paragraph is present for `literature`/`review` but does not conflict with `proposals`. The existing tests (`test_synthesize_stores_in_memory`, `test_synthesize_includes_memory_context`) only check return values and memory storage — they pass regardless of what the system prompt contains.

**`tests/test_followup.py`**: No test verifies the pressure-resistance paragraph is present in `_SYSTEM_PROMPT`. The existing tests mock `_query` entirely, so the system prompt is never exercised.

**Blocker**: Add at minimum:
- A `test_debate.py` test asserting the preserve-disagreement text appears in the `system` string for a `literature` phase call (extend the existing `capture_call` pattern).
- A simple assertion in `test_followup.py` that `"PRESSURE RESISTANCE" in FollowUpChatAgent._SYSTEM_PROMPT` — one line, documents the intent, catches accidental deletion.

---

### Minor

The `base_system` string now ends with `"Be concise and precise.\n\n"` followed by the new IMPORTANT paragraph, forming a double newline before "IMPORTANT". For phases that fall through to `else: system = base_system`, the final system string has a trailing double newline after the paragraph. Not a functional issue, but slightly inconsistent with the single-block style used in `physics_criterion` and `proposals_criterion`.

---

### Verdict: REQUEST_CHANGES

**Two blockers:**

1. The anti-consensus instruction in `base_system` conflicts with the deduplication instruction in `proposals_criterion`. When the model sees both simultaneously for the proposals phase, it receives contradictory guidance: merge overlapping proposals vs. present all views and explain evidential tension. These need to be reconciled so the proposals phase still produces a clean merged list.

2. "Re-evaluate the evidence independently" in `followup.py` should be reworded to anchor the agent to stored analysis results. The current phrasing risks the agent generating new unvalidated reasoning under the guise of independent evaluation, which is the opposite of the conservative epistemic behavior the PR intends to enforce.

Test coverage for the new prompt text is absent and should be added before merge.

PR #26, opened 2026-03-26
fix: run domain classification after file digestion with image context

FaroutYLq:
## Summary

- `_classify_domains()` previously ran **before** files were asked for or digested, so GPD's `route_protocol`/`route_skill` only saw the raw phenomenon text
- Moved the call to **after** `ImageDigestAgent` completes, so `IMAGE_DATA` is already in memory
- The classification input is now enriched with up to 300 chars from each of the first 3 image/document digests (capped at 1000 chars total), giving GPD axis labels, units, data series, and other extracted features to work with

## Test plan

- [ ] Run with a plot whose text description is vague but the image clearly shows a specific domain (e.g. Raman spectrum, SQUID trace) - confirm domain is now detected correctly
- [ ] Run with no files - confirm fallback to phenomenon-only text still works
- [ ] Run existing unit tests: pytest tests/test_memory.py tests/test_debate.py tests/test_phases.py
- [ ] Check DOMAIN_CLASSIFICATION memory entry includes image-informed context when files are supplied

Generated with Claude Code

### DIFF
--- a/mtf/orchestrator.py
+++ b/mtf/orchestrator.py
@@ -53,12 +53,22 @@ class MTFOrchestrator:
         if not self._config.auto_detect_domains:
             return
 
+        # Enrich the classification input with any image/document data already in memory.
+        image_entries = self._memory.filter(MemoryKind.IMAGE_DATA)
+        if image_entries:
+            image_summary = "\n".join(e.content[:300] for e in image_entries[:3])
+            classification_text = f"{phenomenon}\n\n[Extracted file data]\n{image_summary}"
+        else:
+            classification_text = phenomenon
+        # Cap total length sent to GPD
+        classification_text = classification_text[:1000]
+
         detected: list[str] = []
 
         # Try route_protocol (protocols server — always present)
         proto_result = await asyncio.to_thread(
             gpd.call, "protocols", "route_protocol",
-            computation_type=phenomenon[:500],
+            computation_type=classification_text,
         )
         if proto_result:
             normalized = re.sub(r"[\s\-]+", "_", proto_result.lower())
@@ -69,7 +79,7 @@ class MTFOrchestrator:
         # Try route_skill (skills server — gracefully absent if not installed)
         skill_result = await asyncio.to_thread(
             gpd.call, "skills", "route_skill",
-            phenomenon=phenomenon[:500],
+            phenomenon=classification_text,
         )
         if skill_result:
             normalized = re.sub(r"[\s\-]+", "_", skill_result.lower())
@@ -132,10 +142,6 @@ class MTFOrchestrator:
             if gpd is not None and gpd.available:
                 gpd.call("patterns", "seed_patterns")
 
-            # Auto-detect physics domains from the phenomenon (Addition 1)
-            if gpd is not None and gpd.available:
-                await self._classify_domains(phenomenon, gpd)
-
             # If no files were passed programmatically, ask the user interactively
             if files is None:
                 files = await self._interface.ask_for_files() or None
@@ -161,6 +167,10 @@ class MTFOrchestrator:
                     + "\n".join(summary_lines),
                     title="MTF: File Digest",
                 )
+
+            # Auto-detect physics domains — runs after file digestion so IMAGE_DATA is available
+            if gpd is not None and gpd.available:
+                await self._classify_domains(phenomenon, gpd)
 
             # Phase 1: Literature
             hypotheses = await run_literature_phase(


FaroutYLq (commented):
## Review: fix: run domain classification after file digestion with image context

Overall this is a well-motivated and correctly-structured fix. The core idea — delay `_classify_domains()` until after `ImageDigestAgent.digest_all()` has populated `IMAGE_DATA` — is the right solution. Below are my findings across the dimensions requested.

---

### 1. Correctness

The fix solves the stated problem. Previously `_classify_domains()` was called before the file-upload prompt was even shown to the user (`ask_for_files()` came after the old call site), so `IMAGE_DATA` would always be empty. The new ordering is:

```
seed_patterns → ask_for_files → digest_all → _classify_domains → run_literature_phase
```

That sequence ensures `IMAGE_DATA` entries exist when `_classify_domains()` reads `self._memory.filter(MemoryKind.IMAGE_DATA)`. Correctness: PASS.

---

### 2. Edge cases

**Short / single-line digests**: No problem — `e.content[:300]` on a short string just returns the whole string. PASS.

**Empty-string content**: `MemoryEntry.content` is a plain `str` with no enforced minimum length. If `FileDigestSubagent.digest()` ever returns an empty string (e.g., API response with no text block), `e.content[:300]` yields `""` and the join adds a blank line between entries. That is harmless. PASS.

**`image_entries` list is present but all entries are the cross-file synthesis**: When `digest_all()` is called with multiple files it writes N individual `IMAGE_DATA` entries first, then appends a synthesis entry at the end (line 410 of `image_digest.py`). `filter(MemoryKind.IMAGE_DATA)[:3]` therefore picks individual file digests, not the synthesis — which is actually the right choice for domain detection (raw per-file axis labels are more informative than a prose summary). PASS.

**Malformed content (e.g., truncation warning text)**: `digest()` returns the raw LLM text, which can start with headers like `## Image Type\n`. The first 300 chars are still domain-relevant.

---

### 3. The 300/1000 char truncation limits

**300 chars per entry**: The `_IMAGE_SYSTEM_PROMPT` structures output as `## Image Type\n...\n## Axes and Units\n...`. The "Image Type" section is typically 1–2 lines (e.g., "Spectral plot" or "SQUID magnetometry trace"), and "Axes and Units" can begin before 300 chars are consumed. For a Raman spectrum the first 300 chars will almost certainly include `Raman Shift (cm^-1)` and `Intensity (a.u.)` — exactly the tokens GPD needs to route to `condensed_matter/optics`. This limit is appropriate.

**1000 chars total**: The phenomenon text alone was capped at 500 chars in the old code path (`phenomenon[:500]`). After the change, `classification_text` is built as `phenomenon + "\n\n[Extracted file data]\n" + image_summary` and then truncated to 1000. In the worst case — a 500-char phenomenon followed by 3x300 = 900 chars of image data — the total before truncation is ~1415 chars, so the third image digest gets truncated. That is acceptable; the first two digests carry most of the domain signal.

**Minor concern — long phenomenon strings squeeze out image data**: The old code used `phenomenon[:500]` explicitly. The new code removes that cap and relies on the final `[:1000]` alone. If `phenomenon` is, say, 800 chars, the combined string is approximately 800 + 25 (header) + 900 (image) = ~1725 chars, and after `[:1000]` only ~175 chars of image data survive. For very long phenomenon strings the image enrichment is nearly entirely lost. A simple mitigation would be to pre-cap the phenomenon (`phenomenon[:400]`) before building `classification_text`, guaranteeing image data always contributes. This is a minor quality issue, not a correctness bug.

---

### 4. Should `seed_patterns` also move after file digestion?

No. `seed_patterns` calls `gpd.call("patterns", "seed_patterns")`, which seeds the GPD `~/.gpd/` pattern store with built-in patterns. Its purpose is idempotent bootstrapping of the persistent pattern database — it does not read from `SharedMemory` or `IMAGE_DATA` at all. There is therefore no correctness reason to move it after file digestion. Keeping it early is fine.

---

### 5. Other issues

**No-files path is correct**: When the user supplies no files (or answers "no files"), `IMAGE_DATA` will be empty and `_classify_domains` falls back to phenomenon-only text — the same behavior as before the PR. PASS.

**Digestion failure propagation is correct**: If `digester.digest_all()` raises, the exception propagates out of the `if files:` block and `_classify_domains` is never reached (the `finally` branch runs `gpd.close()` instead). No silent partial enrichment. PASS.

**No unit test for the new enrichment path**: The PR test plan lists manual testing only. A unit test that pre-populates `SharedMemory` with a mock `IMAGE_DATA` entry, then calls `_classify_domains` with a mock GPD client, and asserts the routing call receives the enriched text would provide regression protection. This is worth adding before or shortly after merge.

---

### Summary

The fix is correct and the ordering change is the right approach. Two things are worth addressing:

1. **Long-phenomenon truncation**: pre-capping `phenomenon` to ~400 chars before building `classification_text` would guarantee image data always contributes to GPD routing, which is the stated goal of the PR.
2. **Test coverage**: adding a unit test for the enrichment path would prevent regressions.

Neither issue should block merge; both are small follow-ups.

PR #31, opened 2026-03-26
fix: preserve GUI analysis state across browser refreshes

FaroutYLq:
## Summary

- Streamlit `session_state` is wiped on a full page refresh (F5/reload), causing in-progress analyses to appear lost even though the orchestrator thread was still running
- Add a module-level `_live_session` dict that mirrors all mutable orchestrator state (queue, thread, messages, pending, running/finished flags, report, error)
- On each app run, a blank `session_state` is repopulated from `_live_session` so the user seamlessly reconnects to a running or finished analysis

## Behaviour after fix

- All accumulated progress messages are restored after refresh
- Pending `ask`/`confirm` interactions survive refresh — the `reply_q` is preserved so the user can still respond to approval gates
- Final report and error state are restored on completion
- "Start new analysis" clears `_live_session` so the next run starts fresh

## Bug fix (follow-up commit)

If `_live_session` was non-empty but the orchestrator thread died before setting `finished=True` (e.g. event-loop crash before the `try/except` in `_run_orchestrator`), `_initialised` was never set on the session. Neither the inner `if` branch nor the `elif` fired, so the reconnect block re-evaluated on every 0.3s poll rerun indefinitely against a stale dead session. Fixed by moving `ss._initialised = True` outside the inner branch so it always runs when `_live_session` is non-empty.

## Design note

`_live_session` is a module-level singleton — only one analysis at a time across all browser tabs pointing to the same Streamlit server process. Acceptable for this local single-user tool.

## Test plan

- [ ] Start an analysis, wait for at least one progress message, refresh the browser — verify the running view resumes with messages intact
- [ ] Refresh while MTF is waiting at an approval gate — verify the confirm buttons reappear and responding continues the analysis
- [ ] Let an analysis complete, refresh — verify the final report is still shown
- [ ] Click "Start new analysis" after refresh — verify a clean input form appears
- [ ] Kill the orchestrator thread before it sets `finished=True`, refresh — verify the app does not spin in an infinite rerun loop

### DIFF
--- a/mtf/gui.py
+++ b/mtf/gui.py
@@ -11,11 +11,22 @@ import traceback
 from pathlib import Path
 from typing import Any
 
+from mtf.interface import HumanInterface
+
+# ---------------------------------------------------------------------------
+# Module-level session store — survives browser refreshes
+# ---------------------------------------------------------------------------
+# Streamlit session_state is reset on a full page refresh (F5 / reload).
+# We keep a single-slot global mirror of the orchestrator's mutable state so
+# that a refreshed tab can reconnect to a still-running analysis.
+#
+# Keys:  ui_queue, thread, messages, pending, running, finished,
+#        final_report, error, tmp_dir
+_live_session: dict[str, Any] = {}
+
 # ---------------------------------------------------------------------------
 # HumanInterface implementation
 # ---------------------------------------------------------------------------
-
-from mtf.interface import HumanInterface
 
 _REPLY_TIMEOUT = 3600.0  # 1 hour — abandon wait if browser tab was closed
 
@@ -97,6 +108,21 @@ def _streamlit_app() -> None:
     st.caption("Multi-agent AI system for experimental physicists")
 
     ss = st.session_state
+
+    # ------------------------------------------------------------------
+    # Reconnect to a live session after a browser refresh
+    # ------------------------------------------------------------------
+    # On a full page reload st.session_state is blank. If the orchestrator
+    # is still running (or finished) we restore its state so the user can
+    # continue without losing progress.
+    if not ss.get("_initialised") and _live_session:
+        thread: threading.Thread | None = _live_session.get("thread")
+        if thread is not None and (thread.is_alive() or _live_session.get("finished")):
+            for k, v in _live_session.items():
+                setattr(ss, k, v)
+            ss._initialised = True
+    elif not ss.get("_initialised"):
+        ss._initialised = True
 
     # ------------------------------------------------------------------
     # Sidebar — configuration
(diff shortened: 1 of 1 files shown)


FaroutYLq (commented):
## Review

The approach is sound: mirroring Streamlit's ephemeral `session_state` in a module-level dict is the standard pattern for surviving page refreshes, and the plumbing is largely correct. A few issues worth addressing.

---

### Bug: `_initialised` never set when orchestrator thread crashes silently

In the reconnect block (lines 118-125 of the patched file):

```python
if not ss.get("_initialised") and _live_session:
    thread = _live_session.get("thread")
    if thread is not None and (thread.is_alive() or _live_session.get("finished")):
        for k, v in _live_session.items():
            setattr(ss, k, v)
        ss._initialised = True
elif not ss.get("_initialised"):   # only runs when _live_session is EMPTY
    ss._initialised = True
```

If `_live_session` is non-empty **but** the thread has died without ever pushing a `"done"` or `"error"` message (the `try/except` in `_run_orchestrator` covers most cases, but a failure in the event-loop setup before that block could slip through), then:

- the inner `if` is false (thread not alive, `finished` is False)
- the `elif` is also skipped (because `_live_session` is non-empty, so the `elif` condition is never evaluated)
- `ss._initialised` is never set to True

On every subsequent Streamlit rerun the same block re-evaluates identically — effectively an infinite do-nothing loop that re-checks a dead stale session on every poll cycle. Fix: add an `else: ss._initialised = True` inside the outer `if` block to handle the case where `_live_session` is present but the thread condition is not met.

---

### Design limitation: single-slot `_live_session` is not multi-tab safe

`_live_session` is a module-level singleton. If a user opens a second browser tab and starts a new analysis, `_live_session.update(...)` silently overwrites the first run's state. The first tab then "reconnects" to the second run on its next rerun. This is inherent to the single-slot design.

For `mtf-gui` as a local single-user tool this is acceptable today, but it is worth adding a comment to `_live_session` documenting the limitation, so a future contributor does not accidentally make it worse. If multi-tab support is ever needed, keying the store by a run UUID (generated at "Run Analysis" time and persisted across refresh via URL query params or a cookie) would be the straightforward fix.

---

### Minor: `messages` shared-reference comment is misleading

```python
messages=ss.messages,  # shared list reference
```

`ss.messages` is `[]` at this point. The drain loop immediately rebinds `ss.messages` via `ss.get("messages", []) + [new_msg]`, so `_live_session["messages"]` is no longer the same object after the very first "show" message. The comment implies the reference sharing is load-bearing, but it is not — the explicit `_live_session["messages"] = ss.messages` update after each "show" is what actually keeps things in sync. The comment should say something like `# initial empty list; updated explicitly on each message` to avoid misleading future readers.

---

### What is correct

- The `reply_q` object inside `pending` is preserved through `_live_session`, so a refresh during an `ask`/`confirm` gate correctly restores the blocked orchestrator thread's ability to receive the user reply. This is the trickiest part of the PR and it is handled right.
- All `_live_session` mutations happen exclusively in the Streamlit main thread (the drain loop and interaction handlers). The orchestrator thread only writes to `ui_queue`. No locking is needed and none is incorrectly omitted.
- `_live_session.clear()` on "Start new analysis" is correct and complete.
- `tmp_dir` cleanup on "done"/"error" sets `_live_session["tmp_dir"] = None` before the finished state can be read on a future reconnect, so double-cleanup is not possible.

---

**Summary:** The core mechanism is correct and the happy path (refresh during run, refresh at approval gate, refresh after completion) all work. The `_initialised` edge case on silent thread death is a real bug that could cause repeated unnecessary reconnect checks on every poll cycle. The single-slot multi-tab limitation should be documented. Everything else is minor.

